PR #9, opened 2019-05-30
E2E

cnuland:
Fixed E2E tests
Added Helm template test
Removed duplicate files for testing
Added KohlsTechnology quay images URL locations

PR #11, opened 2019-05-30
Issue 3: Add a None option for Resource Handling.

### LINKED ISSUE #3: GitOpsConfig Supports ResourceHandlingMode "None"
## What happens?

I would like `ResourceHandlingMode` to support `None` as an option.

I do not want the `handler` step to do anything. The `process template` step will take care of applying the resources to the k8s cluster.

## Steps to reproduce: 

* N/A

N/A

## Environment:

* Runtime version(Java, Go, Python, etc): N/A
* Desktop OS/version: N/A

## Additional comments:

No.

etsauer:
## Description

Adding a `None` option for resource handling.

Fixes #3 

Must not be merged until tests can be written. Waiting on #9 to add tests.

## Type of change

## Environment

* Runtime version(Java, Go, Python, etc): All

### DIFF
--- a/deploy/crds/eunomia_v1alpha1_gitopsconfig_crd.yaml
+++ b/deploy/crds/eunomia_v1alpha1_gitopsconfig_crd.yaml
@@ -52,19 +52,21 @@ spec:
               type: object
             resourceDeletionMode:
               description: ResourceDeletionMode represents how resource deletion should
-                be handled. Supported values are Retain or Delete. Default is Delete
+                be handled. Supported values are Retain,Delete,None. Default is Delete
               enum:
               - Retain
               - Delete
+              - None
               type: string
             resourceHandlingMode:
               description: ResourceHandlingMode represents how resource creation/update
-                should be handled. Supported values are CreateOrMerge,CreateOrUpdate,Patch.
+                should be handled. Supported values are CreateOrMerge,CreateOrUpdate,Patch,None.
                 Default is CreateOrMerge.
               enum:
               - CreateOrMerge
               - CreateOrUpdate
               - Patch
+              - None
               type: string
             serviceAccountRef:
               description: ServiceAccountRef references to the service account under
--- a/template-processors/base/bin/resourceManager.sh
+++ b/template-processors/base/bin/resourceManager.sh
@@ -53,15 +53,20 @@ function createUpdateResources {
 
 }
 
-echo Managing Resources
+if [ $CREATE_MODE == "None" ]; then
+  echo "CREATE_MODE is set to None; This means that the template processor already applied the resources. Skipping the Manage Resources step."
+  exit 0
+fi
+
+echo "Managing Resources"
 setContext
 
-if [ $ACTION == "create" ] 
+if [ $ACTION == "create" ]
 then
   createUpdateResources
 fi
 
-if [ $ACTION == "delete" ] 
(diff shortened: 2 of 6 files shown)


deploy/crds/eunomia_v1alpha1_gitopsconfig_crd.yaml
```diff
@@ -56,6 +56,7 @@ spec:
               enum:
               - Retain
               - Delete
+              - None
```
seanmalloy: Hmmm ... I think this file is auto generated by the `operator-sdk`, so I think the "correct" way to update this might be to update the Go Lang code then run an `operator-sdk` command to regenerate the file.

But I could be wrong.

raffaelespazzoli: @etsauer you cannot edit a generated file, you need to add the option to the annotation in the `gitopsconfig_types.go` file line: 52. then regenerate everything with:
`operator-sdk generate k8s`
`operator-sdk generate openapi`

etsauer: @raffaelespazzoli yep did that. See latest commit with the updated `gitopsconfig_types.go`, but generation isn't making the changes I expect.

etsauer:
@seanmalloy @raffaelespazzoli I think this one is ready to review.

Smiley73:
Can you please update the readme file with the option and a short explanation what it does?

etsauer:
@Smiley73 done.

etsauer:
Bump

template-processors/base/bin/resourceManager.sh
```diff
@@ -53,15 +53,20 @@ function createUpdateResources {
 
 }
 
-echo Managing Resources
+if [ $CREATE_MODE == "None" ]; then
```
Smiley73: I might be missing something, but how are you handling ResourceDeletionMode = None?

etsauer: @Smiley73 you're right I wasn't. Pushed another commit.

Smiley73: @etsauer of course I was right!
My apologies if I haven't explained this to you previously, so here it is:
1. Heiko is always right
2. You are never right
3. In case you should be right, #1 automatically applies.

etsauer: Thanks for enlightening me.

template-processors/base/bin/resourceManager.sh
```diff
@@ -53,15 +53,20 @@ function createUpdateResources {
 
 }
 
-echo Managing Resources
+if [ $CREATE_MODE == "None" ] || [ $DELETE_MODE == "None" ]; then
+  echo "CREATE_MODE is set to None; This means that the template processor already applied the resources. Skipping the Manage Resources step."
```
Smiley73:
```suggestion
  echo "CREATE_MODE and/or DELETE_MODE is set to None; This means that the template processor manages the resources directly. Skipping the Manage Resources step."
```

Smiley73: @etsauer your force push wiped this change out. I assume that was by accident?

etsauer: sorry, put it back.

PR #51, opened 2019-06-23
Switch to Distroless and Alpine Linux

Smiley73:
## Description

The UBI7 and Fedora images are just huge. Switch to Alpine Linux for the runners and Distroless for the operator, to significantly reduce the size and download/startup times.

## Type of change


sanbornick:
https://github.com/GoogleContainerTools/distroless??

build/Dockerfile
```diff
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-FROM registry.access.redhat.com/ubi7-dev-preview/ubi-minimal:7.6
+FROM alpine:3.10
```
seanmalloy: Let's see if base image for the operator can use [distroless](https://github.com/GoogleContainerTools/distroless) instead of alpine.

Smiley73: I'm pretty sure it can, but I wanted to keep the same base for the runners and the operator itself. This way we only have to maintain one. Thoughts?

seanmalloy: If we are going to switch to a slimmed down based container image then I would prefer to use the distroless base image for the operator image.

I'm fine with having the job pods use a separate base images as those currently require a shell.

Additionally keep in mind that if we switch to a non-RHEL like base image then we cannot get certified. I'm not concerned about this, but just letting you know.

https://github.com/operator-framework/operatorhub-docs/blob/master/distributing-an-operator/distributing-an-operator/certification-requirements.md

Smiley73: Valid point. Not sure how much we care about the certification right now. We can always switch back if we have to.
Can we make the distroless a separate issue? It needs a little bit of testing and I don't think I'll get to that in the next couple of days.

seanmalloy: I think we should just do it right the first time. No need to switch to alpine and then switch to distroless later. Let's take some and test properly instead of rushing through it.

template-processors/base/Dockerfile
```diff
@@ -1,17 +1,18 @@
...
-
-RUN dnf install -y git kubectl gettext python-pip jq findutils && \
-    pip install yq
 
 COPY bin /usr/local/bin
-RUN /usr/local/bin/user_setup
+
+RUN \
+    apk add --no-cache bash curl ca-certificates git gettext jq findutils py-pip && \
+    curl -L https://storage.googleapis.com/kubernetes-release/release/$(curl -s https://storage.googleapis.com/kubernetes-release/release/stable.txt)/bin/linux/amd64/kubectl -o /usr/bin/kubectl && \
```
seanmalloy: I would prefer to pin to a specific version of `kubectl` instead of always using `https://storage.googleapis.com/kubernetes-release/release/stable.txt`.

template-processors/base/Dockerfile
```diff
@@ -1,17 +1,18 @@
...
+    chmod +x /usr/bin/kubectl && \
+    pip install yq && \
```
seanmalloy: We should probably use some sort of version constrain here instead of always pulling the latest version.

PR #57, opened 2019-06-25
Add support for specifying a namespace for the resource to be created in

### LINKED ISSUE #55: Need ability to specify namespace in CR
## What happens?

Currently all resources get created within the namespace the job runs in. This makes it impossible to e.g. create a new namespace, create a new SA and CR within it, then have the operator take it from there.

We can create resources in any namespace, as long as the SA has sufficient access.
We should extend the CRD and add:
```
spec:
  namespace: "xxx"
```
That defaults to the current namespace if it's not defined.
The job templates, and `discoverEnvironment.sh` and `resourceManager.sh` need to set the proper context.

This can simply be passed as an ENV variable to the jobs. The variable already exists, but is not consumed.
```
{{ if .Config.Spec.namespace }}
              value: {{ .Config.Spec.namespace }}
{{ else }}
              valueFrom:
                fieldRef:
                  fieldPath: metadata.namespace          
{{ end }}
```

## Steps to reproduce: 

Try creating a resource in another namespace.

Smiley73:
## Description

Creation of resources in other namespaces was not possible. This PR adds a new optional property `namespace` to be able to specify a specific target namespace.

Fixes #55

## Type of change


deploy/kubernetes/operator.yaml
```diff
@@ -19,7 +19,8 @@ spec:
           image: quay.io/kohlstechnology/eunomia-operator:v0.0.1
           command:
           - eunomia-operator
-          imagePullPolicy: Always
+          #imagePullPolicy: Always
```
seanmalloy: Please delete this commented line.

templates/cronjob.yaml
```diff
@@ -12,12 +12,17 @@ spec:
           containers:
           - name: template-processor
             imagePullPolicy: Always
+            #imagePullPolicy: IfNotPresent
```
seanmalloy: Please remove commented code.

templates/job.yaml
```diff
@@ -11,14 +11,19 @@ spec:
       containers:
       - name: template-processor
         imagePullPolicy: Always
+        #imagePullPolicy: IfNotPresent
```
seanmalloy: Same here. Please remove this.

seanmalloy:
Based on comment https://github.com/KohlsTechnology/eunomia/issues/55#issuecomment-505649886 from @sanbornick we will need to discuss this further before merging.

Smiley73:
I think I had must have had something misconfigured last time I tried this. I got it working now with the existing functionality. Let me close this issue.

PR #59, opened 2019-06-28
Switch to Go modules

### LINKED ISSUE #36: migrate to modules dependency management
currently the project uses `dep`, but the standard for golang going forward is to use modules for dependency management, so we should do this migration. 
documentation and travis CI/CD script should also be updated.

seanmalloy:
## Description

Switch to Go modules.

Using the standard Go Lang process, go mod init, to switch from
dep to modules did not work. Had to run "operator-sdk new foo"
command then copy go.mod and go.sum into this project.

Fixes #36 

## Type of change

## Environment

* Go Lang 1.12.6
* Mac OSX


seanmalloy:
I will remove the project from `snyk.io` and re-add it after merging this PR. This is required when switching Go Lang dependency tools.

seanmalloy:
`snyk.io` do not support Go Modules

PR #67, opened 2019-07-05
switch to helm charts and provide simple gitopsconfig roles

### LINKED ISSUE #48: Role eunomia-operator only works in Eunomia-operator namespace
## What happens?
When using the role eunomia-operator in the demo namespace eunomia-hello-world-demo for the eunomia-operator SA, then the pod doesn't fully initialize.

Setting --zap-level=10 shows the error message.

The eunomia-operator role should work properly.

## Steps to reproduce: 
Deploy container with default role and it will not get past the step "Starting the Cmd."

### LINKED ISSUE #53: Provide cluster roles to manage CRs
## What happens?
It doesn't seem to be sufficient to be an admin of a namespace (at least on OpenShift 3.11) to view the GitOpsConfig CRs.

We should provide roles that will allow viewing and managing CRs. Viewing might be more important, because any of the CRs should be managed via GitOps anyway.

Smiley73:
## Description
- Switch to helm charts to install eunomia and provide simple cluster roles to manage the gitopsconfig objects.
- Helm charts are split into prereqs (requires cluster-admin) and the operator itself (requires namespace admin).
- Updated the cluster demo to leverage new helm charts.

This also fixes:
Fixes #48
Fixes #53

This PR needs further work and is waiting for #9  to be completed, since it will require additional changes for the tests.

## Type of change


DEVELOPMENT.md
```diff
@@ -88,24 +88,26 @@ docker login $REGISTRY
...
 
 Here are some preliminary instructions. This still needs a lot of TLC. Feel free to send in PRs.
 
 ```shell
-oc create namespace eunomia
-oc apply -f ./deploy/crds/gitops_v1alpha1_gitopsconfig_crd.yaml -n eunomia
-oc delete configmap eunomia-templates -n eunomia
-oc create configmap eunomia-templates --from-file=./templates/cronjob.yaml --from-file=./templates/job.yaml -n eunomia
-oc apply -f ./deploy/kubernetes -f ./deploy/openshift -n eunomia
+# Deploy the operator pre-requisites, which require cluster-admin access
```
raffaelespazzoli: it looks like we are creating two charts one for the pre-requisites and one for the actual operator.
I'd recommend creating an wrapper chart that includes those two charts and ensures that the pre-reqs are run first. This will make the user experience easier.
Looking at the pre-reqs charts, I see that it contains also some of the RBAC related manifests. I'm, curious to know the reason of this choice. Usually permissions and roles go in the main chart.

Smiley73: @raffaelespazzoli it's separated by the tasks that need cluster-admin and those that don't. It came from working on the 'cluster' demo. This will allow us to have the operator manage itself, but it will only need admin level access within the namespace.

Smiley73: @raffaelespazzoli I'll have to look into the wrapper chart.

Smiley73: @raffaelespazzoli  I did my required reading... ;-)

We can certainly move this into sub-charts, but I'm not sure how much sense it makes. It would only make it simpler for people testing out the operator, because it's one less command to run (1 instead of 2) to get it installed.
At the end when our users are leveraging our helm charts, they'd most like use them as 2 separate charts due to the required access differences to maintain it.
How do others feel about this? @sanbornick @seanmalloy ?

Smiley73: @raffaelespazzoli @seanmalloy @sanbornick now that the e2e PR is merged, I can finish this one up. What should we do about helm chart and the potential usage of sub-charts. I'm currently not in favor adding sub-charts, because we need to keep it separate for the automation anyway due to different levels of access that are required. The only benefit I see is one less step in the demo.

examples/cluster/README.md
```diff
@@ -51,3 +42,18 @@ At this point the cluster should be "magically" configuring itself and within a
 # Watch the magic happening
 kubectl get pods --all-namespaces -w 
 ```
+
+## Using your own repo
+
+If you would like to test with your own repo, we added a simple overwrite for the main repo URL. This currently only works for the main seed jobs, not the downstream ones for the teams. This is due to the fact that their configuration is entirely loaded from GitHub. You simply need to modify your own repo to point to the correct URLs.
```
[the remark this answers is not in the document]
vinny-sabatini: I dislike super long lines. But it will be displayed the same way

DEVELOPMENT.md
```diff
@@ -88,24 +88,26 @@ docker login $REGISTRY
...
+# Start minikube
 minikube start
-kubectl create namespace eunomia
-kubectl apply -f ./deploy/crds/eunomia_v1alpha1_gitopsconfig_crd.yaml -n eunomia
-kubectl delete configmap eunomia-templates -n eunomia
-kubectl create configmap eunomia-templates --from-file=./templates/cronjob.yaml --from-file=./templates/job.yaml -n eunomia
-kubectl apply -f ./deploy/kubernetes -n eunomia
+
+# Deploy the operator pre-requisites, which require cluster-admin access
+helm template deploy/helm/prereqs/ | kubectl apply -f -
```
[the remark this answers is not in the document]
Smiley73: I believe the latest commit does that. please check.

examples/cluster/teams/platform/cluster-seed/templates/templates/team-seeds.yaml
```diff
@@ -0,0 +1,85 @@
...
+subjects: 
+- kind: ServiceAccount
+  name: {{ .serviceAccountRef }}
+  namespace: {{ .seedNamespace }}
+roleRef:
+  kind: ClusterRole
+  name: admin
+  apiGroup: rbac.authorization.k8s.io  
+---
+# We need to remove this once issue #53 is figured out
```
vinny-sabatini: @Smiley73 is checking if this is needed. Requesting changes until it is figured out

scripts/unit-tests.sh
```diff
@@ -14,8 +14,10 @@
...
 set -e
 
-export JOB_TEMPLATE=$GOPATH/src/github.com/KohlsTechnology/eunomia/templates/job.yaml
-export CRONJOB_TEMPLATE=$GOPATH/src/github.com/KohlsTechnology/eunomia/templates/cronjob.yaml
+export EUNOMIA_PATH=$(cd "${0%/*}/.." ; pwd)
+export JOB_TEMPLATE=${EUNOMIA_PATH}/deploy/helm/operator/eunomia-templates/job.yaml
+export CRONJOB_TEMPLATE=${EUNOMIA_PATH}/deploy/helm/operator/eunomia-templates/cronjob.yaml
 export WATCH_NAMESPACE=""
 export OPERATOR_NAME=eunomia-operator
+
```
vinny-sabatini:
```suggestion
export GO111MODULE=on

```

PR #75, opened 2019-07-30
exit image build script on docker failures

Smiley73:
# Pull Request Template

## Description

The image build script needs to exit out if there's a problem with the image creation.

## Type of change

### DIFF
--- a/scripts/build-images.sh
+++ b/scripts/build-images.sh
@@ -1,4 +1,4 @@
-#!/usr/bin/env bash
+#!/usr/bin/env bash -e
 
 # Copyright 2019 Kohl's Department Stores, Inc.
 #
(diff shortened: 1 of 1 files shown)


scripts/build-images.sh
```diff
@@ -15,28 +15,29 @@
...
 if [ -z "${TRAVIS_TAG}" ] ; then
     IMAGE_TAG="latest"
 else
     IMAGE_TAG=${TRAVIS_TAG}
 fi
 
 # building and pushing the operator images
-docker build . -t ${REPOSITORY}/eunomia-operator:${IMAGE_TAG} -f build/Dockerfile
-docker push ${REPOSITORY}/eunomia-operator:${IMAGE_TAG}
+docker build . -t ${REPOSITORY}/eunomia-operator:${IMAGE_TAG} -f build/Dockerfile || exit 1
```
vinny-sabatini: I think we could change the shebang to `#!/usr/bin/env bash -e` rather than adding `|| exit 1` everywhere

Issue #64, opened 2019-07-03
Cleanup hardcoded v0.0.1 values in repo


vinny-sabatini:
In the Makefile, the version is hard-coded within the LDFLAGS
The template processor images are also hardcoded to build on top of the `v0.0.1` base image.

## What happens?
* `github.com/KohlsTechnology/eunomia/version.Version` is always set to `v0.0.1`
* The template processor images always build on top of the `v0.0.1` image

## What were you expecting to happen?
* `github.com/KohlsTechnology/eunomia/version.Version` is set dynamically 
* The template processor image builds on top of the correct version image
* Correct container image tags are pushed to quay.io when GitHub release is created

vinny-sabatini:
Upon release, this should be easy, just use the new version tag as we do for the image tags.
For `latest` images, I'm not sure the best way to handle this.

Smiley73:
There are a lot of additional places. We really should use `latest` and ask users to pin it internally.

seanmalloy:
Also, we are seeing an issue with the operator container images that are built by Travis CI and push into quay.io.


```
$ oc logs -n eunomia-operator eunomia-operator-6674865f86-j6h4s
standard_init_linux.go:203: exec user process caused "no such file or directory"
```

This build and upload process seems to create usable container images ...
```
$ docker login quay.io
$ ./scripts/build-images.sh quay.io/kohlstechnology
```

Smiley73:
You need to add a `true` ot the build-images script to trigger the upload

PR #92, opened 2019-09-11
Update imagePullPolicy to Always

jkupferer:
So long as the image tag is set to "latest" the imagePullPolicy should be "Always"


seanmalloy:
Hold merging until #64 is fixed, so we don't break the latest tag in quay.io again.

PR #102, opened 2019-09-18
Fix hardcoded v0.0.1 & "latest" tags

### LINKED ISSUE #64: Cleanup hardcoded v0.0.1 values in repo

akavel:
## Description

Fix hardcoded `v0.0.1` and `latest` tags in various places:
- github.com/KohlsTechnology/eunomia/version.Version
- docker image tags on quay.io

As a result of the fix, the tagging policy on quay.io is changed as follows:
- "latest" tag — will be always pointing to the newest official release [Note: this fix will get applied only starting from the next vX.Y.Z release]
- "dev" tag — will be pointing to an image built from the most recently pushed commit in the GitHub repository.

Fixes #64

## Type of change

### DIFF
--- a/scripts/build/get-version.sh
+++ b/scripts/build/get-version.sh
@@ -1,4 +1,5 @@
 #!/usr/bin/env bash
+set -euo pipefail
 
 GIT_V_TAG=$(git tag --list 'v[0-9]*' --points-at HEAD)
 
--- a/scripts/build/get-version.sh
+++ b/scripts/build/get-version.sh
@@ -1,0 +1,10 @@
+#!/usr/bin/env bash
+
+GIT_V_TAG=$(git tag --list 'v[0-9]*' --points-at HEAD)
+
+if [ "$GIT_V_TAG" ]; then
+    echo "$GIT_V_TAG"
+else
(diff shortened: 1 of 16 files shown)


akavel:
Build issues fixed. PTAL.

akavel:
@vinny-sabatini Thanks!

@seanmalloy Would you like to merge this PR?

scripts/build/get-version.sh
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/env bash
+
```
seanmalloy: Please add `set -e` here, so that script will exit if any command fails.

akavel: Done. PTAL

PR #111, opened 2019-10-28
Remove DEFAULT_ROUTE_DOMAIN and REGISTRY_ROUTE environment variables

mkyc:
## Description

This PR removes getDefaultRouteDomain function from discoverEnvironment.sh
script in base template processor because it's not clear why is that
implemented and current implementation fails on kubernetes.

With current implementation hello-world example on Minikube fails with following error: 
```
Cloning Repositories
Cloning into '/git/templates'...
Cloning into '/git/parameters'...
Setting cluster-related environment variable
Context "current" created.
Switched to context "current".
error: the server doesn't have a resource type "route"
Processing Parameters
ERROR - missing parameter files
```

## Type of change

## Environment

* Runtime version(Java, Go, Python, etc): base

### DIFF
--- a/README.md
+++ b/README.md
@@ -186,7 +186,6 @@ The base image provides the following workflow:
     |:---|:---|
     | `CA_BUNDLE`  | Platform-level CA bundle  |
     | `SERVICE_CA_BUNDLE`  | Service-level CA bundle  |
-    | `DEFAULT_ROUTE_DOMAIN`  | Base route domain for the default router  |
     | `NAMESPACE`  | Current namespace  |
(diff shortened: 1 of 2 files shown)


.gitignore
```diff
@@ -79,3 +79,8 @@ vendor
 notes.txt
 image_move.sh
 .DS_Store
+template-processors/applier/Dockerfile
```
Smiley73: Why are we ignoring these Dockerfiles?

mkyc: Those are generated now from Dockerfile.in files in build process.

PR #112, opened 2019-10-29
Removed unused scripts and updated travis ci to run each step once

### LINKED ISSUE #66: Clean Up CI/CD Pipeline
## What happens?

Currently the CI/CD pipeline has various things scattered between .travis.yml, a Makefile, and some shell scripts. It is a bit messy.

A clean easy to follow CI/CD pipeline that is easy to modify.

## Steps to reproduce: 

* n/a

No.

## Environment:

* Runtime version(Java, Go, Python, etc): Go 1.12.6
* Desktop OS/version: n/a

## Additional comments:

I propose that we move everything into the Makefile and remove all of the shell scripts. The Travis CI config would then be updated to just run various `make` commands.

rajneesh-ops:
## Description

This PR removes the unused shell scripts and also updating the existing Travis ci pipeline to run each step once

Fixes #66 

## Type of change

* Bugfix (non-breaking change which fixes an issue)

## Environment


rajneesh-ops:
Suggested changes are done. Please review

.travis.yml
```diff
@@ -31,8 +31,6 @@ before_script:
 
 
 script:
-  - test -z "$(gofmt -l . | grep -v ^vendor)"
```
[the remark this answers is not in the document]
rajneesh-ops: We already have  `go fmt` in e2e-test.sh
image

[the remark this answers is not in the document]
akavel: But running `go fmt ./pkg/... ./cmd...` has the effect of *fixing* the files, locally on the build server only. Whereas in CI, we don't want to actually fix the files; instead, we want to only *check* if the files are correctly formatted. The `test` above did a check; removing it leaves us without a check, no?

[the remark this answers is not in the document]
rajneesh-ops: @seanmalloy shall we keep it? I guess we are executing this test to count the length of the string based on the returned status code.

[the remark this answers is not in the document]
akavel: @rajneesh-ops Did you check the [documentation of `gofmt`](https://golang.org/cmd/gofmt/), including the `-l` flag?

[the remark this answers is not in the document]
seanmalloy: @rajneesh-ops let's also update the `fmt` and `vet` targets in the Makefile, so that we still catch any `gofmt` and `go vet` errors.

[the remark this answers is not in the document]
rajneesh-ops: Ok I am adding the go fmt and go vet then in the Makefile.

rajneesh-ops:
@seanmalloy , @akavel Can you please review the changes

Makefile
```diff
@@ -44,6 +44,10 @@ fmt:
 vet:
 	go vet ./pkg/... ./cmd/...
 
+# Run go fmt Test
+gofmt:
```
akavel: I like the idea of moving the test into Makefile. However, the `gofmt` name above is too similar to l.40, making it hard to understand what's the difference. Seeing this, my proposal is as follows:

1. remove this `gofmt` target, moving its contents (the test line 49) into the `vet` target above
2. move the `golint` call from .travis.yml into the `vet` target above
3. rename the `vet` target to `lint`
    - including all usages of `vet` in the repository

rajneesh-ops: the `vet` is calling at multiple places in Makefile. In that case, we have to update that as well. But the lint will be called multiple times at multiple places.

akavel: Hmm; you're right, this is not so clear cut, for example for quick testing during development we may not want to call full lint (with its complaints about missing godoc comments etc.).

Ok, so I'll back off to just ask you to rename `gofmt` target to `check-gofmt`.

Thanks!

rajneesh-ops: Done

Makefile
```diff
@@ -8,6 +8,7 @@ VERSION := $(shell ./scripts/build/get-version.sh)
 BUILD_COMMIT := $(shell ./scripts/build/get-build-commit.sh)
 BUILD_TIMESTAMP := $(shell ./scripts/build/get-build-timestamp.sh)
 BUILD_HOSTNAME := $(shell ./scripts/build/get-build-hostname.sh)
+GOFMT_OUTPUT := $(shell gofmt -l . | grep -v ^vendor)
```
akavel: why is this in a variable here?  it is used only in one target, and I can't imagine a scenario where we would want to use the feature it adds, of being able to override this variable from outside

rajneesh-ops: Removed please check

Makefile
```diff
@@ -44,6 +44,10 @@ fmt:
 vet:
 	go vet ./pkg/... ./cmd/...
 
+# Run go fmt Test
+check-gofmt:
+	test -z $(shell gofmt -l . | grep -v ^vendor)
```
akavel: please make sure there are double quotes `"` around the `$(...)`.

rajneesh-ops: Done

rajneesh-ops:
Closing this PR and raised a new one
https://github.com/KohlsTechnology/eunomia/pull/136

PR #119, opened 2019-10-30
Correct documentation for running Eunomia locally

### LINKED ISSUE #73: WATCH_NAMESPACE appears to be ignored
It appears that eunomia is ONLY looking in the default namespace and nothing else. This doesn't appear to be affected by setting WATCH_NAMESPACE, either to "" or to a specific namespace.

## What happens?

```
$ echo $WATCH_NAMESPACE
gitops-test
[esauer 🎩︎eunomia] (master)$ operator-sdk up local 
INFO[0000] Running the operator locally.                
INFO[0000] Using namespace default.                     
```

```
$ echo $WATCH_NAMESPACE
gitops-test
[esauer 🎩︎eunomia] (master)$ operator-sdk up local 
INFO[0000] Running the operator locally.                
INFO[0000] Using namespace gitops-test.                     
```

## Steps to reproduce: 

* Follow https://github.com/KohlsTechnology/eunomia/blob/master/examples/hello-world-yaml/README.md

...

## Environment:

* Desktop OS/version:

## Additional comments:

...

sbar95:
## Description

Add --namespace parameter with the value of env variable to
'operator-sdk up local' command. Also add message clarifying the usage
of WATCH_NAMESPACE.

The --namespace parameter of 'operator-skd up local' command always
overrides WATCH_NAMESPACE env variable. If the parameter is not used,
watched namespace equals to "default". Then operator-sdk sets
WATCH_NAMESPACE in Eunomia's environment to that value.

If the parameter is set to empty value (--namespace=""), Eunomia watches
all namespaces.

Fixes #73 

## Type of change

* Bug fix (non-breaking change which fixes an issue

## Environment

* Runtime version(Java, Go, Python, etc): n/a


DEVELOPMENT.md
```diff
@@ -52,9 +52,11 @@ export WATCH_NAMESPACE=""
 export OPERATOR_NAME=eunomia-operator
 export GO111MODULE=on
 go mod vendor
-operator-sdk up local
+operator-sdk up local --namespace="${WATCH_NAMESPACE}"
 ```
 
+Setting WATCH_NAMESPACE to empty string, as above, results in Eunomia watching all namespaces. If you want particular namespace to be watched, set it explicitly in the env variable.
```
seanmalloy: Missing "a" ... should be `If you want a particular`

PR #131, opened 2019-11-04
add fix to set a home directory in the passwd file so that when entry…

### LINKED ISSUE #130: Gitconfig files are not copied to home directory
(While submitting an issue briefly describe the problem you are facing or a new feature you would want to see added) 

## What happens?
I try to set git config files but they do not copy to the home directory
...

The files copy so that auth can be used for a protected git repo
...

## Steps to reproduce: 

* Setup git auth per the readme file
* Create a gitopsconfig and watch the job fail that it cannot find a password

...

## Environment:

* Runtime version(Java, Go, Python, etc): ansible applier
* Desktop OS/version: Mac

## Additional comments:
Have a PR to address the issue
...

mathianasj:
…point runs the home directory is used for the user so that the copy scripts for git config will work

# Pull Request Template

## Description

Add HOME environment variable to the base Docker file so that when the user is added to the passwd file a directory is used that the container will have write access to and thus allow gitconfig files to be copied to appropriate places

Fixes #130

## Type of change

## Environment

* Runtime version(Java, Go, Python, etc): All/base image

### DIFF
--- a/template-processors/base/Dockerfile
+++ b/template-processors/base/Dockerfile
@@ -6,7 +6,8 @@ ENV USER_UID=1001 \
     KUBECTL_VERSION="v1.15.0" \
     YQ_VERSION="2.7.2" \
     GOLANG_YQ_VERSION="2.4.0" \
-    JQ_VERSION="1.6"
+    JQ_VERSION="1.6" \
+    HOME=/tmp
 
 COPY bin /usr/local/bin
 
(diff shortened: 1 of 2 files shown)


seanmalloy:
@mathianasj thanks for opening this pull request!

We will review and get this added to a release milestone in the near future.

seanmalloy:
@mathianasj pull request #195 was just merged. I believe this pull request is no longer required.

akavel:
@mathianasj thanks once again for opening this PR! we went with a slightly different approach in the end, but your PR was very important to us. If you find that you still need something that we overlooked in #195, we'd be super grateful if you could let us know, so we could hopefully try to improve the codebase even more. Thanks, and hopefully see you again in the future! :)

mathianasj:
I like that approach much better, thanks for getting this in

Issue #107, opened 2019-10-09
Error Starting Operator Container


seanmalloy:
## What happens?

Using the `dev` image tag in quay.io we receive the below error when starting the eunomia operator container image.

```
$ kubectl logs -n eunomia-operator eunomia-operator-6674865f86-6l6qv
standard_init_linux.go:203: exec user process caused "no such file or directory"
```

## What were you expecting to happen?

I'm expecting the eunomia operator container to start up without error.

## Steps to reproduce: 

* installer the operator in minikube - https://github.com/KohlsTechnology/eunomia#installing-eunomia
* use the "dev" tag from quay.io instead of the "latest" tag when executing the above install steps
* run `kubectl logs` for the eunomia operator pod to see the error message

## Any errors, stacktrace, logs? 

See error above.

## Environment:

* quay.io `dev` tag

## Additional comments:

n/a

seanmalloy:
Closed by #114

PR #137, opened 2019-11-06
Test dev and latest image tags

shaikmdrafi:
##  Description

It will add the capability to test image 

This commit is added so that we can automatically catch issues like  #107 in the future.
It verifies "dev" tag eunomia operator container image from quay.io can be successfully started in a minikube cluster.

## Type of change

## Environment

### DIFF
--- a/scripts/e2e-test.sh
+++ b/scripts/e2e-test.sh
@@ -47,3 +47,24 @@ operator-sdk test local ./test/e2e --namespace "$TEST_NAMESPACE" --up-local --no
 helm template deploy/helm/eunomia-operator/ \
   --set eunomia.operator.deployment.enabled= \
   --set eunomia.operator.namespace=$TEST_NAMESPACE | kubectl delete -f -
+
+helm template deploy/helm/eunomia-operator/ \
+  --set eunomia.operator.image.tag=dev \
+  --set eunomia.operator.namespace=$TEST_NAMESPACE | kubectl apply -f -
+
+sleep 30
+
+podname=$(kubectl get pods  -o jsonpath='{range .items[*]}{.metadata.name}{"\n"}{end}' -n $TEST_NAMESPACE)
+
+if kubectl exec "${podname}" date -n $TEST_NAMESPACE
+then
+  echo "Pod is Healthy"
+else
+  echo "Pod is not Healthy"
+  exit 1
+fi
+
+helm template deploy/helm/eunomia-operator/ \
+  --set eunomia.operator.image.tag=dev \
+  --set eunomia.operator.namespace=$TEST_NAMESPACE | kubectl delete -f -
+


scripts/test-eunomia.sh
```diff
@@ -0,0 +1,47 @@
+#!/usr/bin/env bash
+set -eu
```
akavel: Would it be ok to change this to `set -euo pipefail`?

scripts/test-eunomia.sh
```diff
@@ -0,0 +1,47 @@
...
+
+deploy_dev_image() {
+	echo "Deploying Dev Image Tag"
+	helm template deploy/helm/eunomia-operator/ | kubectl apply -f -
```
akavel: Why we need this in addition to l.9?

scripts/test-eunomia.sh
```diff
@@ -0,0 +1,47 @@
...
+	echo "Waiting for pod to come up"
+	sleep 30
+}
+
+deploy_latest_image() {
+	printf "\n\n********* Testing Latest image Tag *********\n"
+	helm template deploy/helm/eunomia-operator/ --set eunomia.operator.image.tag=latest | kubectl apply -f -
+	echo "Waiting for pod to come up"
+	sleep 30
+}
```
akavel: why there are so many differences in deploy_latest_image vs. deploy_dev_image? I would expect they both could be nearly identical, what are the reasons they need to be so different?

scripts/test-eunomia.sh
```diff
@@ -0,0 +1,47 @@
...
+
+check_image() {
+	IMAGE_TYPE="$1"
```
akavel: can we use a `local` env variable here?

scripts/test-eunomia.sh
```diff
@@ -0,0 +1,47 @@
+#!/usr/bin/env bash
+set -eu
+
+deploy_dev_image() {
+	echo "Deploying Dev Image Tag"
+	helm template deploy/helm/eunomia-operator/ | kubectl apply -f -
+	sleep 20
+	kubectl config set-context --current --namespace=eunomia-operator
```
akavel: 1. shouldn't we use a test namespace, as in e2e-test.sh?
2. the test doesn't seem to clean up the namespace before starting, so if there is already operator running in it, the test could misleadingly succeed even if kubectl apply below fails, no?
3. also, doesn't `kubectl config set-context` change the namespace permanently for the user of the script, even after the script ends? I don't think we should do something like that, why not modify the namespace only temporarily with helm, as in e2e-test.sh?

akavel:
In PR description and commit message, please add info which issue number this is related to.

scripts/test-eunomia.sh
shaikmdrafi: In deploy_dev_image  function
1) setting namespace  
2) applying all resources in  (deploy/helm/eunomia-operator/)

I believe those are not required  again while using latest tag ,so  ignored .

scripts/test-eunomia.sh
shaikmdrafi: yes we can use, I will change that .

scripts/test-eunomia.sh
akavel: ah; but then we could move those outside the 1st function, as a common preparation for both functions, and then the 2 functions could really be unified to just a 1 function with  1 parameter, no?

scripts/test-eunomia.sh
shaikmdrafi: Its a kind of duplicate , and its fixed

scripts/test-eunomia.sh
shaikmdrafi: Done with suggested changes .

scripts/test-eunomia.sh
shaikmdrafi: done

scripts/e2e-test.sh
```diff
@@ -44,6 +44,8 @@ helm template deploy/helm/eunomia-operator/ \
```
seanmalloy: Can we just add the code from shell function `check_image()` here? Then I think could get rid of the `cripts/test-eunomia.sh` script.

shaikmdrafi: @seanmalloy Removed ` test-eunomia.sh` added pod status check in `e2e-test.sh` itself .

scripts/e2e-test.sh
```diff
@@ -42,6 +42,22 @@ helm template deploy/helm/eunomia-operator/ \
...
+podname=$(kubectl get pods  -o jsonpath='{range .items[*]}{.metadata.name}{"\n"}{end}' -n $TEST_NAMESPACE)
+
+if kubectl exec "${podname}" date -n $TEST_NAMESPACE
+   then
+     echo "Pod is Healthy"
+       else
+     echo "Pod is not Healthy"
+       exit 1
+fi
+
```
seanmalloy: Please move this entire block to the end of the script. I missed this detail in my first review. Sorry.

Also please add the below `helm` command at the very end ...

```
helm template deploy/helm/eunomia-operator/ \
  --set eunomia.operator.namespace=$TEST_NAMESPACE | kubectl delete -f -
```

shaikmdrafi: @seanmalloy Ok i can add `helm` very end . But i have to cleanup(delete)  namespace  again . 
Can we just add After `operator-sdk test local ...` command ?

scripts/e2e-test.sh
seanmalloy: Yes, I believe that will work.

scripts/e2e-test.sh
```diff
@@ -47,3 +47,24 @@ operator-sdk test local ./test/e2e --namespace "$TEST_NAMESPACE" --up-local --no
...
+podname=$(kubectl get pods  -o jsonpath='{range .items[*]}{.metadata.name}{"\n"}{end}' -n $TEST_NAMESPACE)
+
+if kubectl exec "${podname}" date -n $TEST_NAMESPACE
+   then
+     echo "Pod is Healthy"
+       else
+     echo "Pod is not Healthy"
+       exit 1
+fi
```
akavel: Please fix indentation in the `if` block above. Let's keep our code neat and tidy.

scripts/e2e-test.sh
```diff
@@ -47,3 +47,24 @@ operator-sdk test local ./test/e2e --namespace "$TEST_NAMESPACE" --up-local --no
...
+
+helm template deploy/helm/eunomia-operator/ \
+  --set eunomia.operator.deployment.enabled= \
+  --set eunomia.operator.namespace=$TEST_NAMESPACE | kubectl delete -f -
```
akavel: I think we don't need this command, because it is already done in l.47-49; instead, we want the lines that Sean suggested, to delete resources that were created in l.51-53; or am I missing something?

shaikmdrafi: `helm` command in LineNo: 51-53  will recreate all  required NS,SA,CRD's etc.... So just to make sure everything is cleaned properly using delete here.

scripts/e2e-test.sh
shaikmdrafi: Done

scripts/e2e-test.sh
akavel: I completely agree we need to delete the CRDs; but for doing this, how is l.67-69 better than what Sean proposed, i.e.:

```
helm template deploy/helm/eunomia-operator/ \
  --set eunomia.operator.namespace=$TEST_NAMESPACE | kubectl delete -f -
```

?

shaikmdrafi: Just tested and realised proposed `helm` command also deleting all resources that are created .

shaikmdrafi: Done with suggested changes. Kindly review 
Thanks @akavel

PR #149, opened 2019-11-12
deploy/crds: remove resourceHandlingMode=CreateOrUpdate

akavel:
**Description**

Setting `resourceHandlingMode=CreateOrUpdate` was triggering a "kubectl update"
command in `resourceHandler.sh`. However, the "kubectl" command has no
"update" subcommand. This means that this mode is broken. Moreover,
there are no GitHub issues in eunomia mentioning "CreateOrUpdate". This
means nobody uses that mode.

I guess the intention here was probably to use "kubectl replace", not
"kubectl update". However, seeing as it's not actually used, I propose
to remove it and leave only the "kubectl apply" approach. In theory,
it should be enough. If practice tells us that we really need "kubectl
replace", we can introduce it once again, this time with some concrete
test cases and some solid rationale behind the feature.

**Type of change**

* Breaking change (fix or feature that would cause existing functionality to not work as expected)


akavel:
@vinny-sabatini
 
 - as to `zz_generated.deepcopy.go`, it updated after I run `operator-sdk generate k8s`, as suggested here: https://github.com/KohlsTechnology/eunomia/blob/74d13716853ef47c51c42e78dff9c2fd7f2ed8a0/pkg/apis/eunomia/v1alpha1/gitopsconfig_types.go#L38

    This was the ***only*** thing that changed after running this command :/ I was not sure whether to commit it, or not. As of now, I decided to do that basically to show in the commit that this command was run.
 - as to `zz_generated.openapi.go` I changed it ***by hand*** :( I was not able to google up a command that would result in having this file auto-change. Unfortunately, I don't see any `go:generate` statements in the code either, which would help keep a record of what commands we're using to build those autogenerated files :( Maybe @seanmalloy knows how to re-generate `zz_generated.openapi.go`?
 - as to LGTM, can you please hit the "Review Accepted" button to make it more substantial? :) I believe @seanmalloy has the final say on merging or not anyway?

seanmalloy:
Looks good. I think we just need to figure out the correct was to use `operator-sdk` to generate the file `zz_generated.openapi.go`. I'll try to look into this, and see what I find.

akavel:
As was discussed externally, as far as I understand the other command to run is `operator-sdk generate openapi`. With #167 not yet there, we need to run an old version of operator-sdk command; I have a branch nearly ready with a Docker image containing this old version (I didn't invent an easier way of doing this; at least it makes it easy to pin the operator-sdk version), I plan on submitting it as part of this PR.

akavel:
I removed the WIP label and rebased on current master. PTAL

seanmalloy:
@akavel I'll review this one on Monday.

Issue #135, opened 2019-11-05
Add Simple Prometheus Metrics


seanmalloy:
## What happens?

Prometheus metrics are not currently exported by the eunomia operator.

## What were you expecting to happen?

I would like the basic prometheus metrics enabled that are provided by the operator-sdk, so that end users can monitor the health of the eunomia operator.

https://github.com/operator-framework/operator-sdk/blob/v0.8.1/doc/user/metrics/README.md

## Steps to reproduce: 

n/a

## Any errors, stacktrace, logs? 

n/a

## Environment:

Eunomia version v0.0.2.

## Additional comments:

n/a

seanmalloy:
Closed by #150

PR #150, opened 2019-11-13
Add metrics docs

shaikmdrafi:
**Description**

Document the usage of metrics exposed by the operator-sdk, so that end users can monitor the health of the eunomia operator by adding kubernetes-service-endpoints in prometheus scrape_config .

Fixes (#135 )

**Type of change**


README.md
```diff
@@ -254,6 +254,64 @@ helm template deploy/helm/eunomia-operator/ --set openshift.route.enabled=true |
...
+            // handle error
+            log.Info(err.Error())
+        }
+
+        ...
+    }
+```
+
+[scrape_config](https://prometheus.io/docs/prometheus/latest/configuration/configuration/#scrape_config)
+[prometheus-kubernetes](https://github.com/prometheus/prometheus/blob/master/documentation/examples/prometheus-kubernetes.yml)
```
vinny-sabatini: I'm not sure if these are needed since we have the general Prometheus link above.
If we want to keep these, I'd say they should be included in statements explaining the need for them rather than just random links. Maybe something like:
```suggestion
For more details on scrape configurations, see the documentation on [prometheus.io](https://prometheus.io/docs/prometheus/latest/configuration/configuration/#scrape_config).
You can find additional examples on their [GitHub page](https://github.com/prometheus/prometheus/blob/master/documentation/examples/prometheus-kubernetes.yml).
```

README.md
```diff
@@ -254,6 +254,64 @@ helm template deploy/helm/eunomia-operator/ --set openshift.route.enabled=true |
...
+        // Create Service object to expose the metrics port.
+        _, err = metrics.ExposeMetricsPort(ctx, metricsPort)
+        if err != nil {
+            // handle error
+            log.Info(err.Error())
+        }
+
+        ...
+    }
+```
```
vinny-sabatini: I don't believe this block of go code is needed because it's available in the upstream documentation

README.md
```diff
@@ -254,6 +254,64 @@ helm template deploy/helm/eunomia-operator/ --set openshift.route.enabled=true |
...
+
+```
+  - job_name: 'kubernetes-service-endpoints'
+    kubernetes_sd_configs:
+    - role: endpoints
+    relabel_configs:
+      - source_labels: [__meta_kubernetes_namespace]
+        action: keep
+        regex: test-eunomia-operator
+```
```
vinny-sabatini: I would move this up right under the usage header with a short explanation saying what this is.

shaikmdrafi: Done

README.md
shaikmdrafi: Removed

README.md
shaikmdrafi: Done

PR #151, opened 2019-11-13
Add stricter error detection flags in bash scripts

### LINKED ISSUE #74: Eunomia-base image scripts don't handle errors well.
While running through the Hello Worl Yaml sample I get multiple errors. First, it appears the scripts use `find` and `cat`, neither of which exist in the image. Additionally, if the script attempts to delete resources that don't exist, either because its looking in the wrong namespace, or because the resources were already deleted, the script fails, leaving the CR stuck in delete mode as the finalizer was never cleaned up.

## What happens?

Run through the [Hello World YAML](https://github.com/KohlsTechnology/eunomia/tree/master/examples/hello-world-yaml) example. Then manually delete the resources created:

```
kubectl delete deployments.apps/hello-world
kubectl delete svc/hello-world
```

Now delete the cr, and look at the jobs logs from the resulting job.

```
$ kubectl logs gitopsconfig-hello-world-yaml-w4tydf-nfzj4
Cloning Repositories
Cloning into '/git/templates'...
Cloning into '/git/parameters'...
Setting cluster-ralated environment variable
Context "current" created.
Switched to context "current".
error: the server doesn't have a resource type "route"
Processing Templates
Managing Resources
Context "current" modified.
Switched to context "current".
cat: can't open 'find': No such file or directory
No resources found
cat: read error: Is a directory
No resources found
cat: unrecognized option: i
BusyBox v1.30.1 (2019-06-12 17:51:55 UTC) multi-call binary.

Usage: cat [-nbvteA] [FILE]...

Print FILEs to stdout

	-n	Number output lines
	-b	Number nonempty lines
	-v	Show nonprinting characters as ^x or M-x
	-t	...and tabs as ^I
	-e	...and end lines with $
	-A	Same as -vte
No resources found
cat: can't open '.*\.yaml': No such file or directory
No resources found
Error from server (NotFound): error when deleting "/git/manifests/hello-world.yaml": deployments.apps "hello-world" not found
Error from server (NotFound): error when deleting "/git/manifests/hello-world.yaml": services "hello-world" not found
```

...

## Steps to reproduce: 

* ...
* ...

...

## Environment:

* Desktop OS/version:

## Additional comments:

...

akavel:
**Description**

This PR adds error detection flags (`set -euo pipefail`) in all bash scripts I was able to find in the repository, that were missing it before. This will improve robustness of the scripts.

Adding those flags resulted in a failing command being found in `discoverEnvironment.sh`. The PR adds a temporary workaround, and I opened a separate issue #152 with regards to fixing the underlying problem in the future.

Additionally, the PR enables execution tracing (`set -x`) in most of the scripts, including template-processors base scripts. This is intended to help when any debugging of the scripts is needed, and make it easier for contributors to provide better logs when reporting issues.

Closes #74 

**Type of change**

### DIFF
--- a/template-processors/base/bin/resourceManager.sh
+++ b/template-processors/base/bin/resourceManager.sh
@@ -16,39 +16,43 @@
 
 set -euxo pipefail
 
-# this is needed becasue we want the current namespace to be set as default if a namespace is not specified.
+# this is needed because we want the current namespace to be set as default if a namespace is not specified.
 function setContext {
-  $kubectl config set-context current --namespace=$(cat /var/run/secrets/kubernetes.io/serviceaccount/namespace)
+  $kubectl config set-context current --namespace="$(cat /var/run/secrets/kubernetes.io/serviceaccount/namespace)"
   $kubectl config use-context current
 }
 
 function kube {
-  $kubectl -s https://kubernetes.default.svc:443  --token $(cat /var/run/secrets/kubernetes.io/serviceaccount/token) --certificate-authority=/var/run/secrets/kubernetes.io/serviceaccount/ca.crt $@
+  $kubectl \
+    -s https://kubernetes.default.svc:443 \
+    --token "$(cat /var/run/secrets/kubernetes.io/serviceaccount/token)" \
+    --certificate-authority=/var/run/secrets/kubernetes.io/serviceaccount/ca.crt \
+    "$@"
 }
 
 function deleteResources {
-    #first we need to delete the GitOpsConfig resources whose finalizer might not work otherwise
-    for file in $(find $MANIFEST_DIR -iregex '.*\.yaml'); do
-      cat $file | yq 'select(.kind == "GitOpsConfig")' | kube delete -f - --wait=true
+    #first we need to delete the GitOpsConfig resources whose finalizer might not get called otherwise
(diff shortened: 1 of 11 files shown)


template-processors/base/bin/resourceManager.sh
```diff
@@ -14,59 +14,54 @@
...
-    kube patch -R -f $MANIFEST_DIR
-  fi
-
+  case "$CREATE_MODE" in
+    CreateOrMerge)
+      kube apply -R -f "$MANIFEST_DIR"
+      ;;
+    CreateOrUpdate)
+      kube create -R -f "$MANIFEST_DIR"
+      kube update -R -f "$MANIFEST_DIR"
```
mkyc: What about update operation? Did it clarify?

akavel: I'm tracking this in a separate PR #149 so that any one of them does not block the other.

mkyc: cool

seanmalloy:
@akavel looks good. Please resolve the merge conflict, and we can get this merged.

akavel:
@seanmalloy rebased & resolved, PTAL

PR #157, opened 2019-11-15
delete resources removed from repo

### LINKED ISSUE #24: Handle Delete After Push
## What happens?

Removing a yaml file from my git repo does not delete the corresponding k8s object(deployment, service, etc) from my cluster. The object just becomes unmanaged.

After deleting a k8s yaml file from a git repo. I would like this operator to remove that corresponding k8s object(deployment, service, etc) from my cluster.

## Steps to reproduce: 

n/a

no

## Environment:

* Runtime version(Java, Go, Python, etc): n/a
* Desktop OS/version: n/a

## Additional comments:

@raffaelespazzoli has some ideas on how to implement this.

akavel:
**Description**

This PR fixes #24. It uses explicit labels instead of k8s ownership system, as advised by @Smiley73  [on Gitter IIRC?], and following the idea outlined by @raffaelespazzoli.

Quick overview of how it works:
 1. Before applying resources in k8s cluster, the YAMLs are all patched, adding 2 extra labels in their metadata:
    - `owner` — unique ID of the "owning" GitOpsConfig CR
    - `applied` — timestamp when the resource was last applied by eunomia [i.e., resourceManager.sh]
 2. The patched resources (YAMLs) are then `kubectl apply`ed as before
 3. We search for resources with the same `owner`, but with a different `applied` timestamp. This means those resources are "old" — they were present in the repo before, but are not present now, which must mean they were removed. So, we `kubectl delete` them.

Please review; most importantly please let me know if you believe this approach makes sense, or you think there are some problems with it. I'm also open to all implementation & style comments, etc. ~~Before removing the WIP label, I still plan to:~~
- [x] ~~add info in the README~~
- [x] ~~add a test in ./scripts/e2e-test.sh, verifying pods after each step~~
- [x] ~~do one more stylistic pass over the resourceManager.sh script~~
- [x] ~~compose a nice commit message~~

Fixes #24

**Type of change**


seanmalloy:
I think this is a really good start. It looks like there is now a conflict for file `template-processors/base/bin/resourceManager.sh` due to some other recently merged pull requests.

akavel:
@seanmalloy I updated this PR, it now contains the fully developed feature, split into smaller commits and with merge conflicts resolved. Please review, thanks!

akavel:
Please note: I'm now trying to adjust .travis.yml & scripts/e2e-test.sh to set EUNOMIA_URI & _REF to point to the correct repo used in the PR, so that tests would pass in CI. However, the code is otherwise working (the tests pass on my local machine), so I'd be grateful for review of the core changes in the meantime.

akavel:
@seanmalloy @vinny-sabatini This PR now builds and passes tests successfully, so I'd be super grateful if you could spare some time and take a look! Thanks

PR #159, opened 2019-11-15
Add python-passlib to support ansible htpasswd module in applier template processor

pabrahamsson:
**Description**

In order to use the [ansible htpasswd](https://docs.ansible.com/ansible/latest/modules/htpasswd_module.html#requirements) module in the applier template processor image, [PassLib](https://bitbucket.org/ecollins/passlib) has to be installed.

Fixes #(issue)

**Type of change**

### DIFF
--- a/template-processors/applier/Dockerfile.in
+++ b/template-processors/applier/Dockerfile.in
@@ -6,7 +6,7 @@ ENV KUBE_MODULE_VERSION=9.0.0
 USER root
 
 RUN yum install -y --disableplugin=subscription-manager python36-devel gcc python3-pip python3-setuptools && \
-    pip3 install ansible openshift==$OPENSHIFT_MODULE_VERSION kubernetes==$KUBE_MODULE_VERSION jmespath && \
+    pip3 install ansible openshift==$OPENSHIFT_MODULE_VERSION kubernetes==$KUBE_MODULE_VERSION jmespath passlib && \
     yum remove -y gcc python36-devel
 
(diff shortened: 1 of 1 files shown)


akavel:
@seanmalloy what are your thoughts on that? is important enough that we should merge it, or would it be better to try and ask people to write their own template processors based on ours when they need some specific libs for their modules?

@pabrahamsson Thank you a lot for caring enough to submit your first PR to our repo! 😄 ❤️ To better understand your context: have you tried creating your own template-processor image based on `applier`? or do you think it would be too complicated? Personally, it's not clear to me yet where we want to draw the line between stuff that should be included in a template processor image, vs. left to be added by users as a customization. But if you think that creating a separate image is too much of a hassle, maybe we could consider e.g. modifying the entrypoint, to allow customization of imported packages via an ENV variable? What are your thoughts on that?

pabrahamsson:
@akavel @seanmalloy I think it's up to you to decide how you'd like to handle contributions like these to your repository. I have already created my own template-processor image to get me moving while I'm waiting for this PR to merged/closed.
I can see that having to add additional packages in order to support every single ansible module might become a bit of a managing overhead.
If it is feasible to add packages through env vars at container runtime I think that would work for me (and presumably others).

seanmalloy:
I think this is fine. We are not using the `applier` template processor internally at Kohl's, so I don't think I'm going to be too picky about adding additional features.

CC: @etsauer

Issue #101, opened 2019-09-17
Ability to see which commit id was last applied


etsauer:
I would like the ability to see which git commit id was last run. We could add this two places:

1. In the logs of the job pod, both at the beginning and at the end of a run.
2. In the `.status` section of the `GitOpsConfig` resource being acted on.

seanmalloy:
I believe we would want to add this to the `.status` section on the `GitOpsConfig` resources.

seanmalloy:
For the `.status` section in the `GitOpsConfig` resource I'm thinking we would want the below fields.

* a field for the Git SHA1 that is currently being applied
* a field for the last successfully applied git SHA1
* a state field that can have it's value set to Success, InProgress, or Failure that represents the state of the k8s job
* a start timestamp field that represents when the latest k8s job started
* an end timestamp field that represents when the latest k8s job ended

I'd be interested what other people think about the above list of suggested fields.

seanmalloy:
The initial `.status` fields were added in #163, but we still have not added the details git fields to the `.status`.

PR #163, opened 2019-11-19
Add status section for Gitopsconfig CR

safirh:
**Description**
Status section added to the GitOpsConfig, so that end users can view the status of the k8s Jobs that get run when applying config.

Fixes (#101)

**Type of change**


pkg/controller/gitopsconfig/controller.go
```diff
@@ -51,6 +52,13 @@ const kubeGitopsFinalizer string = "eunomia-finalizer"
 // PushEvents channel on which we get the github webhook push events
 var PushEvents = make(chan event.GenericEvent)
 
+//CommitID stores the commitid of template source
+var CommitID string
+
+func init() {
+	CommitID = os.Getenv("TEMPLATE_SHA1")
```
akavel: Did you test that this works? This Go file is part of eunomia operator binary, which AFAIK is normally run in a different Pod than template processors.  I believed that the `template-processors/base/bin/gitClone.sh` script where you seem to be setting TEMPLATE_SHA1 is run as part of template processors, is it not so? If the line above gets non-empty TEMPLATE_SHA1, can you explain to me how? I guess I must be missing something?

pkg/controller/gitopsconfig/controller.go
```diff
@@ -132,6 +141,12 @@ func (r *Reconciler) Reconcile(request reconcile.Request) (reconcile.Result, err
 	}
 	reqLogger.Info("found instance", "instance", instance.GetName())
 
+	//Checking the template commit id
+	CommitID = os.Getenv("TEMPLATE_SHA1")
```
akavel: same question as above

pkg/controller/gitopsconfig/controller.go
```diff
@@ -183,24 +204,40 @@ func (r *Reconciler) createJob(jobtype string, instance *gitopsv1alpha1.GitOpsCo
...
+		log.Info("Error in GetJob:", err)
+		return "", err
+	}
+	if result.Status.StartTime != nil {
+		instance.Status.StartTime = result.Status.StartTime.Format(time.RFC3339)
+		if result.Status.CompletionTime != nil {
+			instance.Status.EndTime = result.Status.CompletionTime.Format(time.RFC3339)
+		}
+	}
+	return getJobStatus(result.Status.Succeeded, result.Status.Failed), nil
```
akavel: This is not a correct way to implement the `state` in GitOpsConfig status field. It is important that  `state` field must be set dynamically, changing quickly in reaction to changes in job status. The code above has the following main issues:

 1. It loads job status into GitOpsConfig status only once, and does not dynamically detect status of the job. For example, if the job takes 90 seconds to finish, the GitOpsConfig status will show `state` "INPROGRESS" even after the job finished.
 2. It blocks the main thread of controller for 5 seconds (l.221) doing nothing. This blocks the controller from working on its main responsibilities, of reconciling GitOpsConfig objects. There must be a better way to dynamically check status of a job than to wait idly.

My suggestion is to find a way to use Kubernetes Go client library to subscribe to job status change events, so that we can be informed about them immediately when they happen, and react by changing the GitOpsConfig status. If that is not possible, we might be able to at least find a way to monitor the job to detect changes in its status.

pkg/controller/gitopsconfig/controller.go
```diff
@@ -239,9 +276,40 @@ func (r *Reconciler) createCronJob(instance *gitopsv1alpha1.GitOpsConfig) error
...
+	if err != nil {
+		log.Info("Error in GetCronJob:", err)
+		return "", err
+	}
+	runningJobArr := result.Status.Active
+	length := len(runningJobArr)
+	if length < 1 {
+		return "FAILED", nil
+	}
+	latestJobRunning := runningJobArr[len(runningJobArr)-1]
```
akavel: The [kubernetes docs for CronJobStatus](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.11/#cronjobstatus-v1beta1-batch) describe `active` field as: "A list of pointers to currently running jobs." I think this is a good idea to check the contents of `active` here. However, I don't understand how do we know that it is enough to check just the last of them?

pkg/controller/gitopsconfig/controller.go
```diff
@@ -417,3 +485,12 @@ func isOwner(owner, owned metav1.Object) bool {
...
 }
+
+func getJobStatus(succeeded, failed int32) string {
+	if succeeded != 0 {
+		return "SUCCESS"
+	} else if failed != 0 {
+		return "FAILED"
+	}
+	return "INPROGRESS"
+}
```
akavel: The general idea here seems to make sense. I would however suggest to improve this function in the following ways:

  1. If possible, to pass whole job status object as argument, and check the Succeeded and Failed fields inside this function, instead of passing 2 ints
  2. If `succeeded + failed > 1`, return error — we expect only 1 pod in the Job as of now, so if there are more of them, this means something is wrong

pkg/controller/gitopsconfig/controller.go
seanmalloy: > My suggestion is to find a way to use Kubernetes Go client library to subscribe to job status change events, so that we can be informed about them immediately when they happen, and react by changing the GitOpsConfig status.

This is called a "watch". Kubernetes client-go or the operator-sdk can create a "watch" to stream k8s changes in real time.

pkg/controller/gitopsconfig/controller.go
```diff
@@ -44,6 +45,13 @@ import (
 )
 
 var log = logf.Log.WithName("controller_gitopsconfig")
+var createJobStartedTime time.Time
```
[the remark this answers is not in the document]
safirh: removed.

pkg/controller/gitopsconfig/controller.go
```diff
@@ -417,3 +439,169 @@ func isOwner(owner, owned metav1.Object) bool {
...
+			instance.Status.EndTime = jobState.EndTime
+			instance.Status.State = jobState.State
+			err := job.updateStatus(instance)
+			if err != nil {
+				log.Info("Error in Updating the job when jobState.Active == 0:" + err.Error())
+			}
+			return
+		}
+
+		if counter == 0 {
```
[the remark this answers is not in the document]
safirh: At very first time it will update the status as "INPROGRESS"... it will check upto job's completion.

pkg/controller/gitopsconfig/controller.go
```diff
@@ -417,3 +439,169 @@ func isOwner(owner, owned metav1.Object) bool {
...
+
+	jobState.State = getJobStatus(jobOutput.Status.Succeeded, jobOutput.Status.Failed, jobOutput.Status.Active)
+	if jobOutput.Status.Active > 0 && jobOutput.Status.Failed > 0 {
+		jobState.Message = "job failed, retrying..."
+	}
+	if jobOutput.Status.StartTime != nil {
+		jobState.StartTime = jobOutput.Status.StartTime.Format(time.RFC3339)
+	}
+	if jobOutput.Status.CompletionTime != nil {
+		jobState.EndTime = jobOutput.Status.CompletionTime.Format(time.RFC3339)
```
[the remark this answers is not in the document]
safirh: changes done

pkg/controller/gitopsconfig/controller.go
```diff
@@ -417,3 +438,169 @@ func isOwner(owner, owned metav1.Object) bool {
...
+		if result.Status.LastScheduleTime != nil {
+			instance.Status.LastScheduleTime = result.Status.LastScheduleTime.Format(time.RFC3339)
+		}
+		instance.Status.Message = "There is no active job"
+		err = job.Reconcile.client.Status().Update(context.TODO(), instance)
+		if err != nil {
+			log.Info("Error in Updating the status for cron job:" + err.Error())
+		}
+	}
+	c := cron.New()
```
akavel: I believe we should add a line above with:

    // TODO: convert from cron to an approach based on Watch, see "Creating watches" in: https://blog.openshift.com/kubernetes-operators-best-practices/

pkg/controller/gitopsconfig/controller.go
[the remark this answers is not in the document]
safirh: We want to set completion time as a `Nil` during first iteration of the loop, that why we need counter. if you have any idea to get the first iteration of the loop.. that will be helpful.

pkg/controller/gitopsconfig/controller.go
```diff
@@ -21,6 +21,7 @@ import (
 	goerrors "errors"
 	"time"
 
+	"gopkg.in/robfig/cron.v2"
```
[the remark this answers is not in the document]
safirh: changes done

pkg/controller/gitopsconfig/controller.go
```diff
@@ -538,70 +543,56 @@ func watchJobForStatus(r *Reconciler, jobName string, namespace string, instance
...
 		log.Info("There is no active job")
-		if result.Status.LastScheduleTime != nil {
-			instance.Status.LastScheduleTime = result.Status.LastScheduleTime.Format(time.RFC3339)
-		}
-		instance.Status.Message = "There is no active job"
-		err = job.Reconcile.client.Status().Update(context.TODO(), instance)
+
+		err = job.updateStatus(instance, gitopsv1alpha1.GitOpsConfigStatus{
+			LastScheduleTime: result.Status.LastScheduleTime,
+			Message:          "There is no active job",
```
akavel: What would you think about adding a line here like below?

    // TODO: set State and other fields based on most recent Job

pkg/controller/gitopsconfig/controller.go
```diff
@@ -538,70 +543,56 @@ func watchJobForStatus(r *Reconciler, jobName string, namespace string, instance
...
 		log.Info("Cron job scheduled, but there is no active job...")
-		if result.Status.LastScheduleTime != nil {
-			instance.Status.LastScheduleTime = result.Status.LastScheduleTime.Format(time.RFC3339)
-		}
-		instance.Status.Message = "Cron job scheduled, but there is no active job..."
-		err = job.Reconcile.client.Status().Update(context.TODO(), &instance)
+
+		err = job.updateStatus(&instance, gitopsv1alpha1.GitOpsConfigStatus{
+			LastScheduleTime: result.Status.LastScheduleTime,
+			Message:          "Cron job scheduled, but there is no active job...",
```
akavel: What would you think about adding a line here like below?

    // TODO: set State and other fields based on most recent Job

pkg/controller/gitopsconfig/controller.go
safirh: i dont think we need to add mentioned line.. because if `active = 0` it has only Last schedule time in job status

pkg/controller/gitopsconfig/controller.go
safirh: i dont think we need to add mentioned line.. because if `active = 0` it has only Last schedule time in job status

pkg/controller/gitopsconfig/controller.go
```diff
@@ -491,43 +487,52 @@ func (job *jobdata) watchjob(instance *gitopsv1alpha1.GitOpsConfig) {
...
 		}
 
 		counter++
 	}
 }
 
-func (job *jobdata) updateStatus(instance *gitopsv1alpha1.GitOpsConfig) error {
-
-	err := job.Reconcile.client.Status().Update(context.TODO(), instance)
+func (job *jobdata) updateStatus(instance *gitopsv1alpha1.GitOpsConfig, jobState gitopsv1alpha1.GitOpsConfigStatus) error {
```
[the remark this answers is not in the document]
safirh: if we go with `instance.Name` and `instance.Namespace` .. getting the below error 
`{"level":"error","ts":1575012770.129529,"logger":"controller_gitopsconfig","msg":"Error in gettting the instance","error":"GitOpsConfig.eunomia.kohls.io \"hello-world-helm\" not found"`.. even though `instance` is a pointer it will be only 4 bytes

pkg/controller/gitopsconfig/controller.go
```diff
@@ -44,6 +45,14 @@ import (
 )
 
 var log = logf.Log.WithName("controller_gitopsconfig")
+var runningJobMap = make(map[string]bool)
```
akavel: Adding a map that is accessed from multiple goroutines adds a dangerous [data race](https://en.wikipedia.org/wiki/Data_race#Software). Please at least change this to a [sync.Map](https://golang.org/pkg/sync/#Map).

pkg/controller/gitopsconfig/controller.go
```diff
@@ -417,3 +437,191 @@ func isOwner(owner, owned metav1.Object) bool {
...
+func (job *jobdata) getCronJobData() (batchv1beta1.CronJob, error) {
+	var result batchv1beta1.CronJob
+	err := job.Reconcile.client.Get(context.TODO(),
+		types.NamespacedName{Name: job.Name, Namespace: job.Namespace},
+		&result)
+	return result, err
+}
+
+func watchJobForStatus(r *Reconciler, jobName string, namespace string, instanceName string, instanceNamespace string, nextSchedule time.Time) {
+	if runningJobMap[jobName] {
```
akavel: as we discussed on chat, I believe we should not do this here; instead, we should rather fix #179 as a completely separate effort. Please remove all usage of runningJobMap in watchJobForStatus.

pkg/controller/gitopsconfig/controller.go
```diff
@@ -417,3 +439,169 @@ func isOwner(owner, owned metav1.Object) bool {
 	}
 	return false
 }
+
+func getJobStatus(succeeded, failed, active int32) string {
```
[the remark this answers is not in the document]
safirh: changes done

pkg/controller/gitopsconfig/controller.go
```diff
@@ -417,3 +439,169 @@ func isOwner(owner, owned metav1.Object) bool {
...
+	if failed != 0 {
+		return "FAILED"
+	}
+	return "INPROGRESS"
+
+}
+
+func (job *jobdata) checkResult() (jobState gitopsv1alpha1.GitOpsConfigStatus, err error) {
+	var jobOutput batchv1.Job
+	err = job.Reconcile.client.Get(context.TODO(),
```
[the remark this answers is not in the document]
safirh: changes done

pkg/controller/gitopsconfig/controller.go
safirh: done

pkg/controller/gitopsconfig/jobmonitor.go
```diff
@@ -0,0 +1,237 @@
...
+func scheduleStatusForCronJobs(job jobmonitor, schedule string, instanceName string, instanceNamespace string) {
+	result, err := job.getCronjobmonitor()
+
+	if err != nil {
+		log.Error(err, "Error in GetCronJob")
+		return
+	}
+
+	//If reconciler calls the same cronjob before the final update of status, we might have an error while
+	//updating the status. To avoid this error runningJobMap is used.
```
akavel: Thank you *very* much for adding those kinds of good comments, explaining the rationale info for *why* some things were done, which is often not easy or even impossible to otherwise find in code. **Highly appreciated!**

pkg/controller/gitopsconfig/jobmonitor.go
```diff
@@ -0,0 +1,237 @@
+package gitopsconfig
```
akavel: I believe there should be a Copyright header in all .go files

safirh: added

seanmalloy:
@safirh the Travic CI tests have failed twice. Please investigate why the tests are failing.

akavel:
Hmm... if increasing Sleep duration in code is the solution to making the tests run, I'm starting to get more doubts about the approach taken in this PR... Most of all, I'm starting to worry, that it may now have bigger impact than I expected, by making *all* tests flaky/non-deterministic, influencing the speed of development for future PRs... @seanmalloy, what is your thinking about this?

akavel:
I added a proposal for a change that seems to fix it on my local machine, see here: https://github.com/safirh/eunomia/pull/1

safirh:
Thanks @akavel ... your changes fixes the travis build error

PR #169, opened 2019-11-23
chore: simplfy Makefile and .travis.yml

seanmalloy:
**Description**

This is the first in a series of commits to clean up and simplify the development work flow. These changes are intended to improve developer velocity.

Removed dead code and unnecessary steps from the Makefile. Also simplify the Travis CI config by using the new "make test" target.

The change is not related to an open issue.

**Type of change**


Makefile
```diff
@@ -15,58 +13,58 @@ export GITHUB_PAGES_REPO ?= KohlsTechnology/eunomia
...
+
 # Install CRDs into a cluster
 install:
 	cat deploy/crds/*crd.yaml | kubectl apply -f-
 
-# Run go fmt against code
+# Run gofmt against code
 fmt:
-	go fmt ./pkg/... ./cmd/...
+	test -z "$(shell gofmt -l . | grep -v ^vendor)"
```
akavel: I'm confused by this block; when running `make fmt` I would be surprised that it's just a *test* instead of actually reformatting my code; also, the comment above says "Run gofmt", but running gofmt would normally reformat my code. Before this change, there was a "check-gofmt" target which did a `test` call, and which was de facto now renamed to "fmt" — why this change in the target's name?

Makefile
```diff
@@ -15,58 +13,58 @@ export GITHUB_PAGES_REPO ?= KohlsTechnology/eunomia
...
-	go build -o build/_output/bin/eunomia  -ldflags $(LDFLAGS) github.com/KohlsTechnology/eunomia/cmd/manager
+clean:
+	rm -rf build/_output
 
-# Build manager binary
-manager-osx: generate fmt vet
-	GOOS=darwin go build -o build/_output/bin/eunomia -ldflags $(LDFLAGS) github.com/KohlsTechnology/eunomia/cmd/manager
+# Build binary
+build:
+	GO111MODULE=on go mod vendor
```
akavel: Why the `go mod vendor`? Why do we need it at all?

Makefile
```diff
@@ -15,58 +13,58 @@ export GITHUB_PAGES_REPO ?= KohlsTechnology/eunomia
...
-manager: generate fmt vet
-	go build -o build/_output/bin/eunomia  -ldflags $(LDFLAGS) github.com/KohlsTechnology/eunomia/cmd/manager
+clean:
+	rm -rf build/_output
 
-# Build manager binary
-manager-osx: generate fmt vet
-	GOOS=darwin go build -o build/_output/bin/eunomia -ldflags $(LDFLAGS) github.com/KohlsTechnology/eunomia/cmd/manager
+# Build binary
+build:
```
akavel: why remove the `generate` target instead of fixing the code by adding the missing directives: `//go:generate operator-sdk generate k8s` and `//go:generate operator-sdk generate openapi` in some good places?

PR #172, opened 2019-11-27
version: add missing build information

seanmalloy:
**Description**

The ldflags being set at build time were not mapping to any variables in
the version package. Added the required variables to the version package to
match what is in the Makefile.

Additionally removed build scripts that are no longer needed.

This pull request is not related to an open GitHub issue.

**Type of change**


Makefile
```diff
@@ -3,20 +3,19 @@
 REGISTRY ?= quay.io
 REPOSITORY ?= $(REGISTRY)/kohlstechnology
 
-BUILD_COMMIT := $(shell ./scripts/build/get-build-commit.sh)
-BUILD_TIMESTAMP := $(shell ./scripts/build/get-build-timestamp.sh)
-BUILD_HOSTNAME := $(shell ./scripts/build/get-build-hostname.sh)
+COMMIT=`git rev-parse HEAD`
+BRANCH=`git symbolic-ref --short -q HEAD || echo HEAD`
+DATE=`date -u +%Y%m%d-%H:%M:%S`
```
sbar95: $() could be used instead of backticks - they are considered [legacy syntax](https://unix.stackexchange.com/a/126928)

Makefile
```diff
@@ -3,20 +3,19 @@
...
+LDFLAG_LOCATION=github.com/KohlsTechnology/eunomia/version
```
akavel: how about calling the variable `VERSION_PKG`  instead, to describe what is the meaning of the value? this is not really a LDFLAG by itself; also, I think we had a similar discussion in a different repo related to Raspberry Pi before ;P

Makefile
```diff
@@ -3,20 +3,19 @@
 REGISTRY ?= quay.io
 REPOSITORY ?= $(REGISTRY)/kohlstechnology
 
-BUILD_COMMIT := $(shell ./scripts/build/get-build-commit.sh)
-BUILD_TIMESTAMP := $(shell ./scripts/build/get-build-timestamp.sh)
-BUILD_HOSTNAME := $(shell ./scripts/build/get-build-hostname.sh)
+COMMIT=`git rev-parse HEAD`
```
akavel: the script had a feature for adding a `-dirty` suffix; would we want to try and keep it?

seanmalloy: My opinion is no I don't think we need to add a `-dirty` suffix. Let me know if you think this would still be a useful feature.

Makefile
seanmalloy: @sbar95 fixed in 8c33c73cbbaf765b26fd30bd0a267caf5e286c7c

Makefile
seanmalloy: @akavel fixed in 928ee721be3f20a1a8b7f4282665e8fee5665643

seanmalloy:
@akavel and @sbar95 please review again, and let me know what you think. I will squash commits before merging.

Makefile
akavel: Hm; now that I think of it, it's too far from the place where it's used anyway, and some commands could be run between the two places, so it could be misleading anyway. So in the end I'm ok with removing it, thanks.

seanmalloy:
I cleaned up the commit history. Should be good for merge now.

PR #198, opened 2019-12-06
template/base: allow hierarchy to include empty directories

### LINKED ISSUE #184: Hierarchy fails on empty directory

eunomia version: 0.0.4

yes

jinja

**What did you do?**
When you have an empty directory as part of the hierarchy is fails on yq merge stating that 2 or more yaml files are needed.

The following line needs to be skipped if no yaml files were found in folder.
https://github.com/KohlsTechnology/eunomia/blob/2e4618d13135de18cb1b4083772f02b1f9aef85c/template-processors/base/bin/processParameters.sh#L54

Empty folders skipped with processing continuing to merge other files found.

**What did you see instead?**
Error where processing stopped.

vinny-sabatini:
**Description**

Allow the hierarchy resolution in the base processParameters.sh script to include directories that do not include any yaml or json files.

In addition, while testing I noticed warnings being logged by the find command due to ordering of options. I have moved the `-name` flags to the end to address this.

Fixes #184 

**Type of change**

### DIFF
--- a/template-processors/base/bin/processParameters.sh
+++ b/template-processors/base/bin/processParameters.sh
@@ -51,7 +51,9 @@ if [ -n "${FOLDERS}" ]; then
     YAML_FILES="$(find "${DIR}" -maxdepth 1 -name \*.json -o -name \*.yaml -o -name \*.yml)"
 
     # merge the files
-    goyq merge -i -x "${VALUES_FILE}" ${YAML_FILES}
+    if [ ! -z "${YAML_FILES}" ]; then
+      goyq merge -i -x "${VALUES_FILE}" ${YAML_FILES}
+    fi
   done
 else
   echo "ERROR - no folders found for processing"
--- a/template-processors/base/bin/processParameters.sh
+++ b/template-processors/base/bin/processParameters.sh
@@ -48,7 +48,7 @@ if [ -n "${FOLDERS}" ]; then
     echo "Processing files in ${DIR}"
 
     # get the list of yaml files to process
-    YAML_FILES="$(find "${DIR}" -name \*.json -o -name \*.yaml -o -name \*.yml  -maxdepth 1)"
+    YAML_FILES="$(find "${DIR}" -maxdepth 1 -name \*.json -o -name \*.yaml -o -name \*.yml)"
 
     # merge the files
     goyq merge -i -x "${VALUES_FILE}" ${YAML_FILES}
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,6 @@
 ## v0.0.6 (Unreleased)
+BUG FIXES:
+* Allow empty hierarchy directories [GH-184](https://github.com/KohlsTechnology/eunomia/pull/184)
 
 ## v0.0.5 (December 5, 2019)
 BUG FIXES:
--- a/test/e2e/testdata/hierarchy/level4/hierarchy.lst
+++ b/test/e2e/testdata/hierarchy/level4/hierarchy.lst
@@ -1,4 +1,5 @@
 ../level1
 ../level2
 ../level3-empty
+../level5-no-files
 ./


template-processors/base/bin/processParameters.sh
```diff
@@ -51,7 +51,9 @@ if [ -n "${FOLDERS}" ]; then
     YAML_FILES="$(find "${DIR}" -maxdepth 1 -name \*.json -o -name \*.yaml -o -name \*.yml)"
 
     # merge the files
-    goyq merge -i -x "${VALUES_FILE}" ${YAML_FILES}
+    if [ ! -z "${YAML_FILES}" ]; then
```
akavel: 1. (Note for future: this can be also shortened to `[ -n "${YAML_FILES}" ]` (as in l. 42), or even `[ "${YAML_FILES}" ]`).
 2. It's not clear to me what's the behavior here with regards to l. 68; do I understand correctly that we'll get this error message now if and only if when there's an empty `hierarchy.lst` file present in the repo?

vinny-sabatini: 1. Good call, thanks for the feedback
2. Looking at the conditional on L.65-70, I believe that is checking if `/tmp/eunomia_values_processed1.yaml` exists, which is created on line 45, so I don't think that error would ever be hit.

The conditional I added will append yaml/json files to `/tmp/eunomia_values_processed1.yaml` if there are any in the current directory being checked, and will move on if there are not any yaml/json files in the directory.

Hope that helps!

akavel: Ad.2: There's a few parts to this story, as I see it:

Firstly, if we are sure this error will never be hit, we should remove it.

However, there's currently one subtle scenario, when I believe ~~it still can be hit~~ there's some tricky behavior: I expect it to be triggered when `hierarchy.lst` file exists, but is empty. In this case, I expect the code flow would go like this:

 1. `FOLDERS` is set to empty on l.21
 2. `if` on l.23 is entered (because `hierarchy.lst` exists)
 3. `while` loop on l.26 is skipped (because `hierarchy.lst` is empty)
 4. `if` on l.42 is skipped (because `FOLDERS` is empty)
 5. ~~`if` on l.65 is skipped (because `VALUES_FILE` is not set) and else on l.67 is entered~~; hm, so, actually, given `set -u` at the beginning of the file, I believe we'll have an error in this case.

Given the above (unless I'm misguided), I believe it would be good if we (1) fix handling of empty `hierarchy.lst` by doing some explicit choice, to make sure it doesn't result in some accidental unexpected code path, and (2) if we want to allow empty/no parameters in the end, prune the dead code fragment on l. 67-69.

Also, notably, I believe our behavior of empty parameters is kinda semi-accidentally drifting. Initially, we seemed to strongly disallow empty params, which was even too strong, as it was breaking hello-world-yaml. However, I guess the idea made some sense for non-trivial template processors (i.e. all except pure `eunomia-base`), to help detect CR errors. Now, after numerous code changes in this file, we seem to have got to a place, where we de facto allow empty parameters; this makes obvious sense for `hello-world-yaml`, but I'm not sure if we are sure we want to remove the protection which existed for non-trivial processors. But if we do, I believe we shouldn't keep the never-used error block in the script, as it might be misleading.

What are your thoughts on this? Would you want to take a look at this by yourself, or would you prefer if I open a new separate issue describing this situation?

Thanks!

vinny-sabatini: Yeah this is a bit tricker than I thought. My vote would be to create a separate issue where we explicitly determine the logic here/how we want to handle empty vars. Then the issue can include tests to enforce this functionality, and if we ever refactor/reduce our bash footprint in this repo, we can make sure we aren't breaking our expected behavior 😄

If you could create the issue with your current understanding of the logic, that would be greatly appreciated!

akavel: Hmm.... so, after some more pondering, I'm starting to think that "tests" is indeed the key word here. I tried to start writing the issue, but without tests, fixing anything here is kinda too fragile anyway; and once we get them, some of this will become much clearer already. So in the end, I think I'll skip for now; it's one of the many papercuts that we'll hopefully have better tools to resolve once we have better tests coverage... Thanks for the discussion! :)

PR #200, opened 2019-12-09
Update docs for user to know which image is used

sbar95:
**Description**

Update documentation is such a way that the user always knows if he's
using his own locally-built image or the one pulled from quay.

This PR doesn't refer to any issue.

**Type of change**

* This change requires a documentation update (it's done in this PR)

### DIFF
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -70,7 +70,7 @@ export GO111MODULE=on
 GOOS=linux make
 ```
 
-From here you could build the eunomia-operator Docker image and manually push it to a registry, or run it locally (out of scope for this doc).
+From here you can build the eunomia-operator Docker image and manually push it to a registry, or run it in a local cluster (see [Using Minikube](#using-minikube) or [Using Openshift](#using-openshift)).
 
 ### Building the image and pushing to a remote registry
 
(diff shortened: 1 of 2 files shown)


DEVELOPMENT.md
```diff
@@ -84,25 +84,54 @@ docker login $REGISTRY
 
 ## Testing
 
-### Using Minikube
+If you want to play with eunomia deployed to Minikube or Minishift, here are some preliminary instructions. They still need a lot of TLC so feel free to send in PRs.
```
akavel: What "TLC" means?

DEVELOPMENT.md
```diff
@@ -84,25 +84,54 @@ docker login $REGISTRY
...
+# Build your eunomia-operator image and store it in minikube's docker registry
+GOOS=linux make
+docker build build/ --tag quay.io/kohlstechnology/eunomia-operator:dev
 
-### Using Openshift
+# Deploy the operator, use your locally-built image
+helm template deploy/helm/eunomia-operator/ \
+  --set eunomia.operator.image.tag=dev \
+  --set eunomia.operator.image.pullPolicy=Never | kubectl apply -f -
+```
```
akavel: Can't we reuse some `make` command here, or something from `scripts/`? Or is there some reason to prefer a list of explicit commands? Looks to me like one more place that we'll forget to change when tweaking something in the build process...

DEVELOPMENT.md
sbar95: Tender Loving Care, I think. Some further improvements for it to always work in this case. Just left the acronym because I think there's always room for improvement.

DEVELOPMENT.md
sbar95: I thought about it but building all images takes much time and this was supposed to be a quick test IMO. We need only eunomia-operator image here to run the operator and `deploy-to-minikube.sh` or `build-images.sh` build all images.
On the other hand now I see an [inconsistency](https://github.com/KohlsTechnology/eunomia/blob/master/DEVELOPMENT.md#building-the-image-and-pushing-to-a-remote-registry). The header says "the image" (also two lines above only eunomia-image is mentioned) but all images are pushed.

DEVELOPMENT.md
akavel: Ohh, ktx. Never knew. TIL, and all those TLAs tend to often be a bit over my head...

DEVELOPMENT.md
akavel: So, who's the intended audience of those instructions?

 - Is it someone who wants to just *use* eunomia? if yes, why choose the `dev` tag?
 - Is it someone who wants to develop on eunomia? if yes, wouldn't they also want to develop the template processors images? At least the `eunomia-base` image is kinda tightly coupled to the eunomia binary itself, we quite often need to change both at the same time...
 - Or is it still someone else?

sbar95: It's the second case - you're right - in most cases `eunomia-base` will be needed too. I'll replace the `docker` command with `GOOS=linux make e2e-test-images` or better remake `deploy-to-minikube.sh` to deploy either to minikube or minishift based on passed argument and use it instead.

Issue #110, opened 2019-10-23
Add Support For k8s Events


seanmalloy:
## What happens?

Resources are applied to my k8s cluster from GitHub, but there are no k8s events created to help track when this happens.

## What were you expecting to happen?

I am expecting k8s events to be created when eunmomia applies k8s resources to a cluster, so that I can easily track when these events occur.

* a k8s event is created when a k8s job fails
* a k8s event is created when a k8s job is successful

The k8s events should contain the k8s job and GitOpsConfig as `InvolvedObjects`. The git SHA1 should be in the `Message`. The `v1.Event` k8s API should be used for creating events.

## Steps to reproduce: 

N/A

## Any errors, stacktrace, logs? 

N/A

## Environment:

* Runtime version(Java, Go, Python, etc):
* Desktop OS/version:

## Additional comments:

...

PR #212, opened 2019-12-17
Emit Kubernetes events when Job completes

akavel:
**Description**

This PR makes eunomia emit Kubernetes events when Jobs it controls
end. This is expected to be useful for tracking when a particular GitOps
change got applied, and whether it applied successfully.

The implementation is based on watching Job resource change events
reported by the Kubernetes client-go API. This PR does not reuse the
mechanisms established in jobmonitor.go. The reasons for this are
multi-fold:
  - This PR is an exploration of the possibility of using Kubernetes
    watching mechanisms (Informers), as a next step after the
    jobmonitor.go code. I believe this approach has a chance to be more
    robust and in-line with the features exposed by the client-go API. I
    hope jobmonitor could be refactored to this same mechanism in future
    as well.
  - Comparing the two approaches, it seems that the one used in this
    PR detects more job completion events than the one used in
    jobmonitor.go.

This PR is a partial fix to #110. However, it does not yet implement
all its requested features:
  - no Git SHA1 in the Message - this was postponed for the time being,
    to reduce the scope of the current task
  - InvolvedObjects field is not used - the Events API as exposed in
    client-go doesn't seem to provide means of setting this field;
    instead, this PR places extra data (Job name) in the Annotations
    field.

Partial fix to #110 

**Type of change**

**Checklist**

- [x] Documentation updated


test/e2e/events_test.go
```diff
@@ -0,0 +1,337 @@
...
+	case event := <-events:
+		if event.Reason != "JobFailed" {
+			t.Errorf("got bad event: %v", event)
+		}
+	case <-time.After(3 * time.Minute):
+		t.Errorf("timeout waiting for JobFailed event")
+	}
+}
+
+func watchEvents(events chan<- *eventv1beta1.Event, timeout time.Duration, namespace, gitops string) (closer func(), err error) {
```
vinny-sabatini: I think this could be moved into `util.go` in case others want to watch for events in their tests

test/e2e/events_test.go
```diff
@@ -0,0 +1,337 @@
...
+	}
+
+	eunomiaURI, found := os.LookupEnv("EUNOMIA_URI")
+	if !found {
+		eunomiaURI = "https://github.com/kohlstechnology/eunomia"
+	}
+	eunomiaRef, found := os.LookupEnv("EUNOMIA_REF")
+	if !found {
+		eunomiaRef = "master"
+	}
```
vinny-sabatini: Probably not a task for this PR, but I think this block could be moved into an "initialize test" function because it seems to be common for every test. @seanmalloy / @akavel if you think this is a good idea I could create the issue for it.

test/e2e/events_test.go
akavel: Done, thanks for the idea!

test/e2e/events_test.go
akavel: I'm totally for it, was thinking about it too, unfortunately just didn't have enough time to fit it in yet. If you could create an issue, that would be awesome; I have some particular ideas how to try and approach this for some extra Go street cred, I'll try to list them as a comment then. Unless I'll find time to do it myself anyway before you open the issue ;P hm, actually, I think I'll try to resist, and leave it as a cookie for grabs for someone else, for bonus Go tutoring value :) unless everybody gets too scared to touch it for fear of an annoying review ;P

test/e2e/events_test.go
akavel: Done, thanks for the idea!

test/e2e/events_test.go
akavel: I'm totally for it, was thinking about it too, unfortunately just didn't have enough time to fit it in yet. If you could create an issue, that would be awesome; I have some particular ideas how to try and approach this for some extra Go street cred, I'll try to list them as a comment then. Unless I'll find time to do it myself anyway before you open the issue ;P hm, actually, I think I'll try to resist, and leave it as a cookie for grabs for someone else, for bonus Go tutoring value :) unless everybody gets too scared to touch it for fear of an annoying review ;P

vinny-sabatini: Well done, it looks like you managed to contain your excitement long enough for me to create #214 😄 Any feedback/suggestions on it are greatly appreciated as always!

PR #234, opened 2020-01-08
Make go test ./... work again

### LINKED ISSUE #126: Standalone `go test ./...` should Just Work™️
Currently, tests require being run with `./scripts/e2e-tests.sh` or `./scripts/unit-tests.sh`. Standalone `go test ./...`, or even `go test ./pkg/...`, doesn't work. 

## What happens?
```
$ go test ./... > /dev/null && echo ok || echo fail
fail
$ go test ./pkg/... > /dev/null && echo ok || echo fail
fail
$ ./scripts/unit-tests.sh > /dev/null && echo ok || echo fail
ok
```

```
$ go test ./... > /dev/null && echo ok || echo fail
ok
```

## Additional comments:

Suggested approach:

 1. Move the env values from `unit-tests.sh` into Go code, possibly to be set via some helper package if they are really needed in more than 1 package in tests. If there is a need to set them to something else somewhere else (in Travis? are they used somewhere else?), provide a way to override them (e.g. via env)
 2. Change the tests in `./test/` to not run by default (for example, by [marking them with a build tag](https://dave.cheney.net/2013/10/12/how-to-use-conditional-compilation-with-the-go-build-tool), e.g. `e2e`), then modify `e2e-tests.sh` to enable running them (for example, by enabling the `e2e` build tag).

sbar95:
**Description**
Replace unit test script with go code and add e2e build tag to end-to-end test files

In init.go add functions that set up environmental variables. Thanks to
this the command "go test -v ./pkg/... works.

The build tag makes "go test ./..." work again. e2e build tag tags the
files to be by default excluded from default tests.

Also, add e2e.go file with package docs to avoid an error of excluding
all go files from e2e package (https://stackoverflow.com/a/47538114).

Initially fixes #126 (I'm going to further research if we could get rid of "operator-sdk test" in favor of "go test")

**Type of change**

**Checklist**

- [ ] Documentation updated (no documentation update necessary for this change)


test/init.go
```diff
@@ -62,3 +67,32 @@ func Initialize() {
...
+func envInit() {
+	logf.Log.Info("Setting up environmental variables")
+	_, initFilename, _, _ := runtime.Caller(0)
+	eunomiaRoot := filepath.Join(filepath.Dir(initFilename), "..")
+	envs := map[string]string{
+		"JOB_TEMPLATE":     filepath.Join(eunomiaRoot, "./build/job-templates/job.yaml"),
+		"CRONJOB_TEMPLATE": filepath.Join(eunomiaRoot, "./build/job-templates/cronjob.yaml"),
+		"WATCH_NAMESPACE":  "",
+		"OPERATOR_NAME":    "eunomia-operator",
+		"GO111MODULE":      "on",
```
akavel: why is this needed? does it even have any effect, setting it when we're already inside `go test`?

test/init.go
```diff
@@ -62,3 +67,32 @@ func Initialize() {
...
+
+// envInit sets up environmental variables for eunomia tests
+func envInit() {
+	logf.Log.Info("Setting up environmental variables")
+	_, initFilename, _, _ := runtime.Caller(0)
+	eunomiaRoot := filepath.Join(filepath.Dir(initFilename), "..")
+	envs := map[string]string{
+		"JOB_TEMPLATE":     filepath.Join(eunomiaRoot, "./build/job-templates/job.yaml"),
+		"CRONJOB_TEMPLATE": filepath.Join(eunomiaRoot, "./build/job-templates/cronjob.yaml"),
+		"WATCH_NAMESPACE":  "",
```
akavel: Why is this needed?

test/init.go
```diff
@@ -62,3 +67,32 @@ func Initialize() {
...
 }
+
+// envInit sets up environmental variables for eunomia tests
+func envInit() {
+	logf.Log.Info("Setting up environmental variables")
+	_, initFilename, _, _ := runtime.Caller(0)
+	eunomiaRoot := filepath.Join(filepath.Dir(initFilename), "..")
+	envs := map[string]string{
+		"JOB_TEMPLATE":     filepath.Join(eunomiaRoot, "./build/job-templates/job.yaml"),
+		"CRONJOB_TEMPLATE": filepath.Join(eunomiaRoot, "./build/job-templates/cronjob.yaml"),
```
akavel: Wouldn't it be easier to pass the paths as regular arguments to the InitializeTemplates function on l.68? Is there some important reason why we must pass them through environment instead?

test/init.go
```diff
@@ -62,3 +67,32 @@ func Initialize() {
...
+		"WATCH_NAMESPACE":  "",
+		"OPERATOR_NAME":    "eunomia-operator",
```
akavel: Does something break if we remove this?

akavel:
Thanks for the copyright headers, and for making the code so much simpler after the review!

I think after the post-review changes, the commit message fragment: "`add functions that set up environmental variables`" no longer really makes much sense? Other than that, LGTM!

seanmalloy:
@sbar95 looks good to me. Please adjust the commit message as mentioned by @akavel. After that I will merge this pull request. Thanks!

PR #237, opened 2020-01-10
fix for race condition between the jobs

### LINKED ISSUE #179: Race condition between jobs started by eunomia

eunomia version: 0.0.4

yes

$ kubectl version
Client Version: version.Info{Major:"1", Minor:"14", GitVersion:"v1.14.8", GitCommit:"211047e9a1922595eaa3a1127ed365e9299a6c23", GitTreeState:"clean", BuildDate:"2019-10-15T12:11:03Z", GoVersion:"go1.12.10", Compiler:"gc", Platform:"darwin/amd64"}
Server Version: version.Info{Major:"1", Minor:"16", GitVersion:"v1.16.2", GitCommit:"c97fe5036ef3df2967d086711e6c0c405941e14b", GitTreeState:"clean", BuildDate:"2019-10-15T19:09:08Z", GoVersion:"go1.12.10", Compiler:"gc", Platform:"linux/amd64"}
</pre></details>

**What did you do?**

As noticed by @safirh, there is race condition between jobs started by eunomia. Specifically, if two reconciliation requests for a CR (say, "A" and "B") are submitted after each other (esp. when the time between them is short; for example, as a result of two Webhook triggers from two commits), eunomia would *start* a new Job for the 2nd version of the CR (B) *before the Job for the 1st one (A) ended*. The two jobs would then be running concurrently, and with nondeterministic scheduling in Kubernetes, it can happen that the Job B will end (run resourceManager.sh) *before* Job A. As a result, Job A's "apply" would then override Job B's "apply", leaving cluster in old state.

This gets even more problematic when we implement a "status" field for the CRD in eunomia, as adding job info and git SHA1 info there would be confusing and misleading, with races between jobs running in parallel.

I would expect that new job, for applying a newer version of a CR, would be postponed until the previous job (for an older version of the CR) finished (with success or failure).

**What did you see instead?**

Jobs for two versions of a CR from the same Git repository may run concurrently, leading to a race condition, where the result of applying the older version/commit of the CR may override the result of applying the newer version/commit of the CR.

safirh:
**Description**
The new job for applying a newer version of a CR, would be postponed until the previous job get completed
Fixes #179 

**Type of change**

**Checklist**

- [ ] Documentation updated


pkg/controller/gitopsconfig/controller.go
```diff
@@ -207,31 +208,51 @@ func ContainsTrigger(instance *gitopsv1alpha1.GitOpsConfig, triggeType string) b
...
+
+	for _, jobFromList := range jobList.Items {
+		if isOwner(instance, &jobFromList.ObjectMeta) {
+			if jobFromList.Status.Active != 0 {
+				log.Info("Job is runnning for this instance...." + instance.Name)
+				return reconcile.Result{Requeue: true}, nil
+			}
+		}
+
+	}
```
akavel: Please now try to write an e2e test for this feature.

My initial idea how I would try to write the test, is:
 1. trigger eunomia multiple times quickly, for example, by [sending values to gitopsconfig.PushEvents channel in a similar way as is done in eunomia Webhook handler code](https://github.com/KohlsTechnology/eunomia/blob/e1900b95cf469cfeb45f748da9bf773ebb54b641/pkg/handler/handler.go#L100-L103)
      - if this doesn't work initially, it may be because of rate limiting in operator-sdk; if yes, it may require finding rate limiting options in operator-sdk and tweaking them
 2. verify somehow in logs that message from line 231 above was emitted, meaning previous job was detected and new one postponed
 3. verify that after first job finished, second one (that was postponed) becomes started

pkg/controller/gitopsconfig/controller.go
```diff
@@ -207,31 +208,51 @@ func ContainsTrigger(instance *gitopsv1alpha1.GitOpsConfig, triggeType string) b
...
+	}, jobList)
+
+	if err != nil {
+		log.Error(err, "unable to list the jobs")
+		return reconcile.Result{}, err
+	}
+
+	for _, jobFromList := range jobList.Items {
+		if isOwner(instance, &jobFromList.ObjectMeta) {
+			if jobFromList.Status.Active != 0 {
```
akavel: Please merge this `if` condition with the one above, like this:

    if isOwner(...) && jobFromList.Status.Active != 0 {

akavel:
@safirh How about [my comment on .ObjectMeta](https://github.com/KohlsTechnology/eunomia/pull/237#discussion_r365241071)? It doesn't seem done, and you didn't reply with an explanation why it doesn't work, so I don't know if you missed it, or there are reasons why it can't be done?

pkg/controller/gitopsconfig/controller.go
safirh: unable to fetch/read the operator logs to run the e2e test.

seanmalloy: I discussed with with @akavel in chat ... we are not sure the best way to write unit or e2e tests for this code change, so that this time we are merging without tests being added.

seanmalloy:
@safirh I think the code changes look good. One suggestion for the future would be to use `git rebase` to clean up commit history.

PR #245, opened 2020-01-13
chore: enable code coverage using codecov.io

seanmalloy:
**Description**

Enabling unit test code coverage tracking using https://codecov.io.

This pull request is not related to an open issue.

**Type of change**

**Checklist**

- [x] Documentation updated

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -47,6 +47,9 @@ before_script:
 script:
   - make test-dirty test
 
+after_success:
+  - bash <(curl -s https://codecov.io/bash)
+
 deploy:
   - provider: script
     skip_cleanup: true
--- a/Makefile
+++ b/Makefile
@@ -60,7 +60,7 @@ test-e2e:
 
 .PHONY: test-unit
 test-unit:
-	go test -v -coverprofile=coverage.txt ./pkg/...
+	go test -v -coverprofile=coverage.txt ./...
 
 # Install CRDs into a cluster
 .PHONY: install
--- a/Makefile
+++ b/Makefile
@@ -60,7 +60,7 @@ test-e2e:
 
 .PHONY: test-unit
 test-unit:
-	go test -v ./pkg/...
+	go test -v -coverprofile=coverage.txt ./pkg/...
 
 # Install CRDs into a cluster
 .PHONY: install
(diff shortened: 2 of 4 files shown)


Makefile
```diff
@@ -60,7 +60,7 @@ test-e2e:
 
 .PHONY: test-unit
 test-unit:
-	go test -v ./pkg/...
+	go test -v -coverprofile=coverage.txt -covermode=atomic ./pkg/...
```
akavel: Why `-covermode atomic`? The `go help testflag` manual seems to imply that by default some reasonable values are used for this flag; is it an explicit requirement of codecov.io? If yes, please add a comment line above l.63 with an explanation.

.travis.yml
```diff
@@ -47,6 +47,9 @@ before_script:
 script:
   - make test-dirty test
 
+after_success:
+  - bash <(curl -s https://codecov.io/bash)
```
akavel: Oooo, I love how simple it is to use

seanmalloy:
@akavel here is what it will look like ... https://codecov.io/gh/KohlsTechnology/git2consul-go. Once this PR is merged then future pull requests will report on the code coverage. The report can't be generated automatically because we have not uploaded any data for the master branch yet.

Makefile
seanmalloy: I followed the Go example for codecover.io. This is where I got `covermode atomic` from.

https://github.com/codecov/example-go#codecov-go-example

It looks like they recommend `atomic` or `count`. Let me now which you prefer.

seanmalloy: @akavel if you have some time to let me know what you think about this. That would be great. I'd ideally like to merge this tomorrow.

seanmalloy:
I also added commit 56c1e6c09636e05d46d60652eeb70d17c307a02d which runs the unit tests for all `.go` source files in the project instead of just files in the `pkg/` directory.

akavel:
Sorry for being late with getting back to this.

Actually, I'm not really convinced we need the `-covermode` flag. In the codecov link you shared (thanks!), it seems to me to be marked only as a "note", not a hard requirement. I'm not sure there's some important value in knowing how many times a line was executed; I don't have experience with code coverage, but on first look, I'm much more interested in *which* lines were executed *at all*; more than that, I think I actually find the execution counts only distracting. So, if you're ok with that, I'd prefer if we tried to see if it works with no `-covermode`.

Other than that, LGTM.

seanmalloy:
I've updated the `go test` command to not include the `-covermode` CLI option. We will see what happens.

Issue #217, opened 2019-12-19
Refactor GitOpsConfig Status Update Code


seanmalloy:
**Is your feature request related to a problem? Please describe.**

No, this is a code refactoring enhancement.

**Describe the solution you'd like**

I would like the code(`pkg/controller/gitopsconfig/jobmonitor.go`) that updates the `.status` section of the GitOpsConfig CR to use the same approach(k8s Informers) as Event creation.

**Describe alternatives you've considered**

The current code that updates the `.status` section is not as reliable as we would like.

**What version of eunomia are you using?**

eunomia version: v0.0.5

**Additional context**

We believe using k8s Informers for updating the `.status` section will be more robust than than current approach. See PR #212.

akavel:
When completing this work, please remove the [relevant TODO in jobmonitor.go](https://github.com/KohlsTechnology/eunomia/blob/525d288c85377b4a84a9011696a78276ba50b9e8/pkg/controller/gitopsconfig/jobmonitor.go#L33).

Also related: #179.

seanmalloy:
I'm closing this for now. We can improve the code that updates the `.status` section in the future. For now I believe it is working much better than before. Closed by #249.

PR #242, opened 2020-01-12
add proper error check for jobs not found

Smiley73:
**Description**
Cronjobs that were deleted are now properly detected and Eunomia does not end up in an endless loop.

Fixes #241 

Replace the unreliable string comparison for "not found" error checks, with the proper k8s function.

```
{"level":"info","ts":1578841086.369399,"logger":"gitopsconfig-controller","msg":"retrying the GetJob for CheckResult"}
{"level":"info","ts":1578841086.369462,"logger":"gitopsconfig-controller","msg":"Retry of GetJob for CheckResult unsuccessful: Job.batch \"gitopsconfig-hello-world-helm-cfsv6c\" not found"}
{"level":"info","ts":1578841086.369468,"logger":"gitopsconfig-controller","msg":"The job has been deleted"}
```

**Type of change**
* Bug fix (non-breaking change which fixes an issue)

**Checklist**

- [ ] Unit tests and e2e tests updated
- [ ] Documentation updated

PR #249, opened 2020-01-14
pkg/controller/gitopsconfig: update Status based on informers (#217)

akavel:
**Description**

This PR changes the mechanism for updating the Status field of
GitOpsConfig from cron (pull-based) to informers (push-based).

The implementation in this PR is unfortunately not ideal. I
encountered many challenges when developing it, and in the end took
compromises in some aspects. Specifically:

1. Status is now calculated based on single Job's data received from
   Informer. As such, this makes Status calculations "edge-based",
   instead of "level-based", with the latter being the recommended
   approach for Kubernetes apps. In case Informer misses some changes
   (for example, potentially in case of lost network connection), we may
   miss recalculating some GitOpsConfig Status transition.

   I tried changing this approach to "level-based". In OnUpdate, I tried
   List-ing all Jobs belonging to GitOpsConfig, and calculating Status
   based on the newest one among them. However, to my surprise, List was
   giving me *stale data* from Kubernetes. I considered adding an extra
   cron-like task, that would periodically recalculate Status for all
   GitOpsConfig CRs; however, it's not yet clear to me how to implement
   it in a stateless and robust way in eunomia.

   In the end, I reverted back to the "edge-based" approach as shown in
   this PR, finding it the simplest implementation for now, while
   seemingly (to my surprise) giving robust results. I am considering a
   theory, that maybe the Informers infrastructure is robust enough that
   it takes some complexity from its users.

2. Setting GitOpsConfig Status can currently fail in case the CR was
   modified between the Get() and Status().Update() calls. Current
   implementation doesn't do any retries in such case, thus making it
   possible for Status to become stale and incorrect.

   However, it's not clear to me how to implement retries correctly.
   I expect retries shouldn't block the OnUpdate function with
   Sleep-like mechanism, as that would risk stalling that part of
   eunomia. Also, I am not sure how to do it in a stateless way, such
   that it would retry also in case eunomia is restarted in meantime.

   Again, I would consider trying to use a cron-like mechanism here,
   possibly as complementary to "fast" edge-based updates. However, I
   see challenges in how to implement such cron well in a stateless and
   scaling-safe way, so I decided to not do it yet, and bring it up to
   discussion.

As a potential workaround, if nobody comes up with a better idea, I
would like to propose adding a background thread in eunomia, repeatedly
picking a random GitOpsConfig from the cluster, and recalculating its
Status. That could be a method to achieve eventual consistency, though
in a non-deterministic way.

Additionally, through the course of implementing this feature, I came to
believe it would be good to simplify tracking the relation between
GitOpsConfig, and Jobs it spawns, by marking the Jobs with a new label,
with GitOpsConfig name as its value. This would simplify the "owner
detection" code in numerous places in eunomia, currently done in
numerous different ways, and also make it simpler and more future-proof.

I wanted to make it part of the work done in this PR, but I decided
to postpone it for now, as implementing this feature as is was already
taking me too long.

Partially implements #217

**Type of change**

**Checklist**

- [ ] Documentation updated


Smiley73:
@akavel I like this being stateless and some remediation task running in the background to "self-heal" anything that got out of whack. 
Wouldn't we need that logic anyway if the pod get restarted? We need to expect Eunomia to go down at any point in time and then being able to pick up where it left off.
Worst case you could leverage configmaps to write out some kind of status to get a bit more stateful, but I'm not a huge fan of it. It might be better to have a shared filesystem between the pods to accomplish this.

Smiley73:
Also if this doesn't get merged before the release, then I please need #242 merged in the interim.

seanmalloy:
Looks pretty good. I believe this much closer to the correct way for updating `.status` than what we had before.

But I see that `go vet` is failing. Can you please address these issues listed by `go vet` from the Travis CI run ...

https://travis-ci.com/KohlsTechnology/eunomia/builds/144404755#L771-L784

akavel:
Hmh, sorry for not noticing :/ I don't seem to be able to find the magic incantation to make Travis send build statuses to my work email :/

akavel:
Now the actual test is failing; I'll be working on it, may be submitting some WIP commits.

akavel:
Should hopefully be fixed now... let's see what Mr. Travis has to say on that...

PR #250, opened 2020-01-14
Remove e2e-test.sh script

### LINKED ISSUE #126: Standalone `go test ./...` should Just Work™️

sbar95:
**Description**
Replace e2e test script by go code so that `go test ./test/e2e/... -tags e2e` executes e2e tests.

Fixes no issue as for now - if finished this code is to fix #126 

**Type of change**

**Checklist**

- [ ] Documentation updated (no need here)

### DIFF
--- a/scripts/build-images.sh
+++ b/scripts/build-images.sh
@@ -29,7 +29,11 @@ PUSH_IMAGES=${2:+true}
 # Usage: build_image <context dir> <image name>
 # Example: build_image template-processors/myimage/ myimage
 build_image() {
-  local context_dir=$1
+  local eunomia_root=$(cd $(dirname $0)/..; pwd)
+  case $1 in
+    /*) local context_dir=$1 ;; # already absolute path
+    *)  local context_dir="$eunomia_root/$1" # use absolute path
+  esac
   local image_url=${REPOSITORY}/$2
   local push=${PUSH_IMAGES:-false}
 
--- a/scripts/build-images.sh
+++ b/scripts/build-images.sh
@@ -29,10 +29,9 @@ PUSH_IMAGES=${2:+true}
 # Usage: build_image <context dir> <image name>
 # Example: build_image template-processors/myimage/ myimage
 build_image() {
-  local eunomia_root=$(cd $(dirname $0)/..; pwd)
   case $1 in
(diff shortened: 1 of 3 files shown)


sbar95:
@seanmalloy @akavel 
Regarding our recent discussion: the code is not finished, some things are to be added yet. I wanted to share it to show you the general direction where my researches to replace `operator-sdk test local ...` with `go test ...` are going. Some things I wasn't able to run using go APIs, so I used `exec` which is poor IMO.

Now looking at this I am of the opinion that we should rather stick to `operator-sdk`. Rewriting it to Go generates too much code to maintain (and at this point it's a rather poor one) - if we stick to the foregoing approach we have only several dozen lines of bash code which is more natural than rewriting it to Go at a push.

seanmalloy:
My opinion is that we should just stick with using the operator-sdk to run e2e tests. So, we would not merge this pull request.

@akavel do you have any thoughts on this?

seanmalloy:
Discussed with @akavel and @sbar95 and at this time we believe that using the e2e test functionality from `operator-sdk` is the best approach.

PR #253, opened 2020-01-14
[WIP]Include hello-world-yaml example in e2e testing

### LINKED ISSUE #133: Run all examples as part of real e2e tests
`./scripts/e2e-test.sh` should run all scenarios from `./examples/`; current e2e-test.sh has way too small coverage of the code, didn't catch a lot of bugs we had, and many new developments are still not covered by it.

## What happens?

Running `./scripts/e2e-test.sh` did not tell us about #115 — all tests were passing, though delete functionality was broken. The example scenarios include running the delete functionality, but e2e-test.sh was passing successfully when delete was broken.

Also, `./examples/hello-world-yaml` was failing previously on my dev machine, whereas e2e-test.sh still was passing successfully.

- [x] `./scripts/e2e-test.sh` should run all scenarios from `./examples/`
    - for example, it should be failing when hello-world-yaml is failing
- [ ] when any examples are changed (including changes in the commands in `./examples/*/README.md`), the e2e-test.sh should automatically update (for example, by automatically extracting the commands to run from the README.md files)
- [x] `./scripts/e2e-test.sh` must then pass in Travis-CI

## Steps to reproduce: 

* change something in `./examples/hello-world-helm/` (or `-yaml/`) to make it broken
* run hello-world-helm (or -yaml) example to verify it is broken
* run `./scripts/e2e-test.sh` — observe that it still reports success

shaikmdrafi:
**Description**
Ensure that all examples are run as part of the e2e tests .

Fixes #133 

**Type of change**

**Checklist**

- [ ] Documentation updated

### DIFF
--- a/test/e2e/yaml_test.go
+++ b/test/e2e/yaml_test.go
@@ -1,0 +1,96 @@
+// +build e2e
+
+/*
+Copyright 2020 Kohl's Department Stores, Inc.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+	http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package e2e
+
+import (
+	goctx "context"
+	"os"
+	"testing"
+
+	framework "github.com/operator-framework/operator-sdk/pkg/test"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	"github.com/KohlsTechnology/eunomia/pkg/apis"
+	gitopsv1alpha1 "github.com/KohlsTechnology/eunomia/pkg/apis/eunomia/v1alpha1"
+)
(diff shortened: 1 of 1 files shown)


akavel:
It is not clear to me if this PR in its current state is a good way to approach #133. 

As much as I think it may be possible to resolve #133 through a Go test in `test/e2e/` somehow, my current  belief is that it's a harder and more tricky way, if at all possible. In fact, as of yet it is not clear to me  how to solve #133 using a Go test in `test/e2e/`, and as such I find it difficult to provide guidance on how to to approach solving this that way. For an alternative solution I would find easier to understand, please read below.

To clarify, my current understanding of #133 is such, that we would like to eventually have automatic tests verifying, as much as possible, that the example instructions we give to users through READMEs in `examples/` directory are correct and working. I may be missing some ideas or a different interpretation here, especially for purposes of this PR; please help me understand if that is so. I don't understand how this PR in its current shape is aiming to get us closer to this goal.

My current belief is, that the simplest and easiest to maintain way to achieve the goal described above, would be to try and copy as many as possible instructions from each `examples/*/README.md` file into a new function in the `e2e-test.sh` script, then make a minimal amount of changes required to make them run successfully as part of e2e-test.sh on local machines and in CI. That's an approach I would personally suggest to take when trying to solve #133. As a first step, I agree it makes sense to me to do it for just a single example, e.g. for `examples/hello-world-yaml/README.md` (this is the example that I see being mentioned in the code of this PR).

PR #257, opened 2020-01-17
Use golang.org/x/xerrors instead of raw `return err`

### LINKED ISSUE #178: Start using golang.org/x/xerrors instead of raw `return err`

I am wary of the way how errors are currently returned from functions in the codebase — for example, code blocks like below:

```go
if err != nil {
    log.Error(err, "...<some message>...", "foo", bar)
    return err
}
```

As much as I understand the value of structured logs, I'm especially uncomfortable with the "naked" `return err`. This isn't a good practice in Go, in part because it loses information for tracking the details of the problem at intermediate steps, when the error finally surfaces somewhere higher.

I suggest using https://golang.org/x/xerrors package, and through its [Errorf](https://godoc.org/golang.org/x/xerrors#Errorf) function, adding helpful context information when returning an error, instead of a raw `return err`.

I considered also removing `log.Error` calls; but I think those can be still useful for getting information in a more structured way (JSON), better for bulk log analysis than the format used in xerrors. 

eunomia version: 0.0.4 / dev

**Additional context**

karolkrawczyk:
**Description**

Use golang.org/x/errors so errors have a better context for
easier troubleshooting.

Fixes #178 

**Type of change**

**Checklist**

- [not needed] Unit tests and e2e tests updated
- [not needed] Documentation updated

### DIFF
--- a/pkg/controller/gitopsconfig/controller.go
+++ b/pkg/controller/gitopsconfig/controller.go
@@ -19,6 +19,7 @@ package gitopsconfig
 import (
 	"context"
 	goerrors "errors"
+	"golang.org/x/xerrors"
 	"time"
 
 	batchv1 "k8s.io/api/batch/v1"
@@ -76,7 +77,7 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
 	// Create a new controller
 	c, err := controller.New(controllerName, mgr, controller.Options{Reconciler: r})
 	if err != nil {
-		return err
+		return xerrors.Errorf("failed creation of controller %q: %w", c, err)
 	}
 
 	// Watch for changes to primary resource GitOpsConfig
@@ -101,7 +102,7 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
 		},
 	)
 	if err != nil {
-		return err
+		return xerrors.Errorf("controller watch for changes to primary resource GitOpsConfig failed: %w", err)
 	}
 
 	err = c.Watch(
@@ -109,7 +110,7 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
 		&handler.EnqueueRequestForObject{},
 	)
 	if err != nil {
-		return err
+		return xerrors.Errorf("controller watch for PushEvents failed: %w", err)
 	}
 
 	// TODO: we should somehow detect when Reconciler is stopped, and run the
@@ -120,7 +121,7 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
 		eventRecorder: mgr.GetRecorder(controllerName),
 	})
 	if err != nil {
-		return err
+		return xerrors.Errorf("cannot create watch job for jobCompletionEmitter handler: %w", err)
 	}
 
 	// TODO: detect when Reconciler stops and run stop func returned by addJobWatch
@@ -129,7 +130,7 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
 		client: mgr.GetClient(),
 	})
 	if err != nil {
-		return err
+		return xerrors.Errorf("cannot create watch job for statusUpdater handler: %w", err)
 	}
 
 	return nil
@@ -170,7 +171,7 @@ func (r *Reconciler) Reconcile(request reconcile.Request) (reconcile.Result, err
 			return reconcile.Result{}, nil
 		}
 		// Error reading the object - requeue the request.
-		return reconcile.Result{}, err
+		return reconcile.Result{}, xerrors.Errorf("cannot read kubernetes object: %w", err)
 	}
 	reqLogger.Info("found instance", "instance", instance.GetName())
 
@@ -199,8 +200,9 @@ func (r *Reconciler) Reconcile(request reconcile.Request) (reconcile.Result, err
 		reconcileResult, err := r.createJob("create", instance)
 		if err != nil {
 			reqLogger.Error(err, "error creating the job, continuing...")
+			return reconcileResult, xerrors.Errorf("error creating the job: %w", err)
 		}
-		return reconcileResult, err
+		return reconcileResult, nil
 	}
 
 	return reconcile.Result{}, err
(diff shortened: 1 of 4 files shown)


pkg/controller/gitopsconfig/controller.go
```diff
@@ -244,19 +246,19 @@ func (r *Reconciler) createJob(jobtype string, instance *gitopsv1alpha1.GitOpsCo
 	job, err := util.CreateJob(mergedata)
 	if err != nil {
 		log.Error(err, "unable to create job manifest from merge data", "mergedata", mergedata)
-		return reconcile.Result{}, err
+		return reconcile.Result{}, xerrors.Errorf("unable to create job manifest from merge data %q: %w", mergedata, err)
```
akavel: I believe mergedata may be quite huge, so I'm not convinced we want to try printing it all here; also, what will `%q` print in this case?

pkg/controller/gitopsconfig/controller.go
```diff
@@ -244,19 +246,19 @@ func (r *Reconciler) createJob(jobtype string, instance *gitopsv1alpha1.GitOpsCo
...
 	}
 	err = controllerutil.SetControllerReference(instance, &job, r.scheme)
 	if err != nil {
 		log.Error(err, "unable to the owner for job", "job", job)
-		return reconcile.Result{}, err
+		return reconcile.Result{}, xerrors.Errorf("unable to the owner for job %q: %w", job, err)
```
akavel: 1. the sentence doesn't seem to gramatically make sense to me; can you try to understand or ask what the original author meant here, and improve this message in both lines above?
2. I am not sure %q will print something sensible for whole "job" object
3. In a similar vein, please review all the remaining  uses of `%q` in this PR, to make sure they actually print something sensible and useful, while also not dumping whole huge object; please let me know afterwards, I'll come back to reviewing this PR then

pkg/controller/gitopsconfig/controller.go
```diff
@@ -282,14 +284,14 @@ func (r *Reconciler) createCronJob(instance *gitopsv1alpha1.GitOpsConfig) error
 			update = false
 		} else {
 			// Error reading the object - requeue the request.
-			return err
+			return xerrors.Errorf("cannot read kubernetes object: %w", err)
```
akavel: please try to add more context to this message; see also comment on l.174

pkg/controller/gitopsconfig/controller.go
```diff
@@ -282,14 +284,14 @@ func (r *Reconciler) createCronJob(instance *gitopsv1alpha1.GitOpsConfig) error
...
 		}
 	}
 
 	err = controllerutil.SetControllerReference(instance, &cronjob, r.scheme)
 	if err != nil {
 		log.Error(err, "unable to the owner for cronjob", "cronjob", cronjob)
-		return err
+		return xerrors.Errorf("unable to the owner for cronjob %q: %w", cronjob, err)
```
akavel: please improve as mentioned above for similar line

pkg/controller/gitopsconfig/controller.go
```diff
@@ -320,7 +322,7 @@ func (r *Reconciler) GetAll() (gitopsv1alpha1.GitOpsConfigList, error) {
 	err := r.client.List(context.TODO(), &client.ListOptions{}, instanceList)
 	if err != nil {
 		log.Error(err, "unable to get the list of GitOpsCionfig")
-		return *instanceList, err
+		return *instanceList, xerrors.Errorf("unable to get the list of GitOpsCionfig: %w", err)
```
akavel: please fix typo: Cionfig -> Config

pkg/util/util.go
```diff
@@ -56,14 +57,14 @@ func InitializeTemplates(jobTemplateFileName string, cronJobTemplateFileName str
 
 	jobTemplate, err = jobTemplate.Parse(string(text))
 	if err != nil {
-		log.Error(err, "Error parsing template", "template", text)
-		return err
+		log.Error(err, "error parsing template", "template", text)
+		return xerrors.Errorf("error parsing template %q: %w", text, err)
```
akavel: please don't dump whole template contents into error message

pkg/controller/gitopsconfig/controller.go
karolkrawczyk: Done.

pkg/controller/gitopsconfig/controller.go
karolkrawczyk: Fixed, please let me know if new error message sounds better.

pkg/controller/gitopsconfig/controller.go
karolkrawczyk: Sure.

pkg/controller/gitopsconfig/controller.go
```diff
@@ -19,6 +19,7 @@ package gitopsconfig
 import (
 	"context"
 	goerrors "errors"
+	"golang.org/x/xerrors"
```
akavel: package xerrors is not in Go standard library, so it should go in the block between l.25-l.42, here and in all files below

seanmalloy:
Looks pretty good. Please just address the final comment from @akavel https://github.com/KohlsTechnology/eunomia/pull/257#discussion_r368627912

karolkrawczyk:
@akavel I believe I fixed import order. Please review once again when you'll have some spare time, thank you.

pkg/controller/gitopsconfig/controller.go
```diff
@@ -391,17 +393,17 @@ func (r *Reconciler) manageDeletion(instance *gitopsv1alpha1.GitOpsConfig) (reco
...
 	// looking up all delete jobs
 	err = r.client.List(context.TODO(), &client.ListOptions{
 		Namespace:     instance.GetNamespace(),
 		LabelSelector: selector,
 	}, jobList)
 	if err != nil {
-		log.Error(err, "unable to list jobs ")
-		return reconcile.Result{}, err
+		log.Error(err, "unable to list all delete jobs")
+		return reconcile.Result{}, xerrors.Errorf("unable to list all delete jobs: %w", err)
```
akavel: on l. 180, there's an implicit "return err"; as a result, if this error happens, we won't know which instance it's related to; please either add this info in this error message, or expand l.180 to add it

akavel: after discussion, can be left as is, thanks!

PR #258, opened 2020-01-17
Include examples in e2e testing

### LINKED ISSUE #133: Run all examples as part of real e2e tests

shaikmdrafi:
**Description**

Ensure  examples are tested as part of the e2e tests .

Fixes #133 

**Type of change**

**Checklist**


examples/hello-world-yaml/test.yaml
```diff
@@ -0,0 +1,44 @@
...
+    ref: master
+    contextDir: examples/hello-world-yaml/parameters
+  triggers:
+  - type: Change
+  #- type: Periodic
+    #cron: '*/1 * * * *'
+  serviceAccountRef: eunomia-runner-yaml
+  templateProcessorImage: quay.io/kohlstechnology/eunomia-base:latest
+  resourceHandlingMode: Apply
+  resourceDeletionMode: Delete
```
akavel: why copy whole contents of already existing files into another file? why not run the same commands as listed in examples/hello-world-yaml/README.md?

scripts/e2e-test.sh
```diff
@@ -78,3 +78,24 @@ helm template deploy/helm/eunomia-operator/ \
...
   --set eunomia.operator.namespace=$TEST_NAMESPACE | kubectl delete -f -
 
+helm template deploy/helm/eunomia-operator/ \
+  --set eunomia.operator.image.tag=dev \
+  --set eunomia.operator.image.pullPolicy=Never \
+  --set eunomia.openshift.route.enabled=true | kubectl apply -f -
+
+kubectl apply -f examples/hello-world-yaml/test.yaml
+
+sleep 30s
```
akavel: don't use sleep, try using `kubectl wait` similar as in l.65 instead

scripts/e2e-test.sh
```diff
@@ -78,3 +78,24 @@ helm template deploy/helm/eunomia-operator/ \
...
+
+RESULT=`kubectl get pods -n eunomia-hello-world-yaml-demo -l name=hello-world -o=jsonpath="{range .items[*]}{.status.phase}{'\n'}{end}"`
+
+if [[ $RESULT == "Running" ]]
+then
+        echo "Example Passed"
+else
+	echo "Example Failed"
+        exit 1
+fi
```
akavel: please fix indentation in above block

scripts/e2e-test.sh
```diff
@@ -78,3 +78,24 @@ helm template deploy/helm/eunomia-operator/ \
   --set eunomia.operator.image.tag=dev \
   --set eunomia.operator.namespace=$TEST_NAMESPACE | kubectl delete -f -
 
+helm template deploy/helm/eunomia-operator/ \
```
akavel: please add a comment above explaining that below block is testing the hello-world-yaml example

scripts/e2e-test.sh
shaikmdrafi: Actually i tried  to use wait flag. 
But what's happening is CR is creating deployment, if i use wait flag its checking before CR is creating deployment and its failing with no deployment found. 
I believe  is wait  flag will work only  for existing resource at that point , in my case CR is taking 20s to become success . 
I tried using wait for GitOpsConfig to become ready . But it did not worked seems GitOpsConfig don't have that field to check wait flag.

examples/hello-world-yaml/test.yaml
Smiley73: The whole purpose of this exercise is to use the actual examples. We need to know if any of them stop working due to changes, which will gives us a nice functional coverage on our e2e tests.

shaikmdrafi: Ok Got it. Removed file.

scripts/e2e-test.sh
```diff
@@ -78,3 +78,29 @@ helm template deploy/helm/eunomia-operator/ \
   --set eunomia.operator.image.tag=dev \
   --set eunomia.operator.namespace=$TEST_NAMESPACE | kubectl delete -f -
 
+# Below block is to ensure hello-world-yaml example is working
+helm template deploy/helm/eunomia-operator/ \
+  --set eunomia.operator.image.tag=dev \
+  --set eunomia.operator.image.pullPolicy=Never \
+  --set eunomia.openshift.route.enabled=true | kubectl apply -f -
```
akavel: why enable openshift route?

scripts/e2e-test.sh
```diff
@@ -78,3 +78,29 @@ helm template deploy/helm/eunomia-operator/ \
...
+
+kubectl apply -f examples/hello-world-yaml/cr/hello-world-cr1.yaml -n eunomia-hello-world-yaml-demo
+
+sleep 30s
+
+RESULT=`kubectl get po -n eunomia-hello-world-yaml-demo -l name=hello-world -o=jsonpath="{range .items[*]}{.status.phase}{'\n'}{end}"`
+
+if [[ $RESULT == "Running" ]]
+then
+  echo "Example Passed"
```
akavel: in the hello-world-yaml example, there is also ...cr2.yaml and ...cr3.yaml, why are you not checking those, but you say "Example Passed"?

scripts/e2e-test.sh
shaikmdrafi: Thanks @akavel .
Removed sleep condition.

scripts/e2e-test.sh
shaikmdrafi: Fixed .

scripts/e2e-test.sh
shaikmdrafi: Ok, i will add steps validate cr2 and cr3 as well 👍

scripts/e2e-test.sh
```diff
@@ -78,3 +78,28 @@ helm template deploy/helm/eunomia-operator/ \
...
+kubectl apply -f examples/hello-world-yaml/cr/hello-world-cr1.yaml -n eunomia-hello-world-yaml-demo
+
+timeout=30
+
+while ((--timeout)) && [[ "$(kubectl get po -n eunomia-hello-world-yaml-demo -l name=hello-world -o=jsonpath="{range .items[*]}{.status.phase}{'\n'}{end}")" != "Running"  ]];do
+  echo "waiting for hello-world deployment: remaining $timeout sec..."
+  sleep 1
+done
+if [[ $timeout == 0 ]]; then
+  echo "Example CR1 Test FAILED"
```
akavel: please write full name of the example also in this message ("hello-world-yaml-demo") - we plan to test other examples in future too, so this will be helpful to know which one of them is failed

scripts/e2e-test.sh
```diff
@@ -78,3 +78,28 @@ helm template deploy/helm/eunomia-operator/ \
   --set eunomia.operator.image.tag=dev \
   --set eunomia.operator.namespace=$TEST_NAMESPACE | kubectl delete -f -
 
+# Below block is to ensure hello-world-yaml example is working
+helm template deploy/helm/eunomia-operator/ \
+  --set eunomia.operator.image.tag=dev \
+  --set eunomia.operator.image.pullPolicy=Never | kubectl apply -f -
+
+kubectl create namespace eunomia-hello-world-yaml-demo
```
Smiley73: It might be a good idea to adding an `echo` before each command, in case it fails. Otherwise it'll be harder to debug.
E.g. `echo creating namespace xxx`

akavel: The script has `set -x` at the beginning, so I think it should be enough?

Smiley73: Good point.

scripts/e2e-test.sh
```diff
@@ -87,19 +87,53 @@ kubectl create namespace eunomia-hello-world-yaml-demo
...
+    exit 1
+  fi
+  echo "Example hello-world-yaml-cr1 Test Passed"
+}
+
+#Testing hello_world_yaml_cr_2
+hello_world_yaml_cr_2() {
+  kubectl apply -f examples/hello-world-yaml/cr/hello-world-cr2.yaml -n eunomia-hello-world-yaml-demo
+  while ((--timeout)) && [[ "$(kubectl get replicaset -n eunomia-hello-world-yaml-demo -l name=hello-world -o=jsonpath="{range .items[*]}{.status.readyReplicas}{'\n'}{end}")" != "3"  ]];do
+    echo "waiting for hello-world-yaml-cr1 deployment: remaining $timeout sec..."
```
akavel: please fix `cr1` to `cr2`

scripts/e2e-test.sh
```diff
@@ -87,19 +87,53 @@ kubectl create namespace eunomia-hello-world-yaml-demo
...
+    echo "Example hello-world-yaml-cr2 Test FAILED"
+    exit 1
+  fi
+  echo "Example hello-world-yaml-cr2 Test Passed"
+}
+
+#Testing hello_world_yaml_cr_3
+hello_world_yaml_cr_3() {
+  kubectl apply -f examples/hello-world-yaml/cr/hello-world-cr3.yaml -n eunomia-hello-world-yaml-demo
+  while ((--timeout)) && [[ "$(kubectl get deployment -n eunomia-hello-world-yaml-demo -o=jsonpath="{.items[0].metadata.generation}{'\n'}")" != "3"  ]];do
```
akavel: why `.items[0]`? previous checks have `{range .items[*]}`, why different here?

shaikmdrafi: Both works , Will keep unique .

scripts/e2e-test.sh
shaikmdrafi: fixed

scripts/e2e-test.sh
```diff
@@ -87,19 +87,53 @@ kubectl create namespace eunomia-hello-world-yaml-demo
...
-echo "Example CR1 Test Passed"
-
+timeout=45
+
+#Testing hello-workld-yaml-cr1
+hello_world_yaml_cr_1() {
+  kubectl apply -f examples/hello-world-yaml/cr/hello-world-cr1.yaml -n eunomia-hello-world-yaml-demo
+  while ((--timeout)) && [[ "$(kubectl get po -n eunomia-hello-world-yaml-demo -l name=hello-world -o=jsonpath="{range .items[*]}{.status.phase}{'\n'}{end}")" != "Running"  ]];do
+    echo "waiting for hello-world-yaml-cr1 deployment: remaining $timeout sec..."
+    sleep 2
```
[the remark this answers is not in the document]
shaikmdrafi: Thanks I did not noticed that .
Increased timeout, and using sleep 1 only.

scripts/e2e-test.sh
akavel: I don't know kubectl very well: can you help me understand why cr1 and cr2 use `range` with `items[*]`, but cr3 uses `items[*]` without `range`?

shaikmdrafi: It's just used to filter json output . 
cr3  passing values are available in 2 places  in json output, so i used first field. 
Updated now for keeping unique.

shaikmdrafi:
> One major thing remaining about this PR: in the PR description, it is written "fixes #133". But in description of #133, there is written "run all examples", and this PR runs only one example, hello-world-yaml. There are [more examples in eunomia](https://github.com/KohlsTechnology/eunomia/tree/1345a37fed68aefe42e01f1ecebc92d58304526d/examples) than just hello-world-yaml. In my current understanding, I would expect to either:
> 
> * a) keep the description "fixes #133", and include all remaining examples also in the PR
> * b) keep only hello-world-yaml in the PR, but fix the description to not say "fixes #133", maybe instead say "partial fix for #133"
> 
> What are your thoughts about this?

I was testing only this  hello-world-yaml example bcs , i believe other example are being tested via operrator-sdk

shaikmdrafi:
> > One major thing remaining about this PR: in the PR description, it is written "fixes #133". But in description of #133, there is written "run all examples", and this PR runs only one example, hello-world-yaml. There are [more examples in eunomia](https://github.com/KohlsTechnology/eunomia/tree/1345a37fed68aefe42e01f1ecebc92d58304526d/examples) than just hello-world-yaml. In my current understanding, I would expect to either:
> > 
> > * a) keep the description "fixes #133", and include all remaining examples also in the PR
> > * b) keep only hello-world-yaml in the PR, but fix the description to not say "fixes #133", maybe instead say "partial fix for #133"
> > 
> > What are your thoughts about this?
> 
> I was testing only this hello-world-yaml example bcs , i believe other example are being tested via operrator-sdk

But  seems Test's are not similar to what is mentioned in README.md

scripts/e2e-test.sh
```diff
@@ -78,3 +78,62 @@ helm template deploy/helm/eunomia-operator/ \
...
+  --set eunomia.operator.image.tag=dev \
+  --set eunomia.operator.image.pullPolicy=Never | kubectl apply -f -
+
+kubectl create namespace eunomia-hello-world-yaml-demo
+
+kubectl apply -f examples/hello-world-yaml/eunomia-runner-sa.yaml -n eunomia-hello-world-yaml-demo
+
+timeout=90
+
+#Testing hello-workld-yaml-cr1
```
vinny-sabatini:
```suggestion
#Testing hello-world-yaml-cr1
```

shaikmdrafi:
> > > One major thing remaining about this PR: in the PR description, it is written "fixes #133". But in description of #133, there is written "run all examples", and this PR runs only one example, hello-world-yaml. There are [more examples in eunomia](https://github.com/KohlsTechnology/eunomia/tree/1345a37fed68aefe42e01f1ecebc92d58304526d/examples) than just hello-world-yaml. In my current understanding, I would expect to either:
> > > 
> > > * a) keep the description "fixes #133", and include all remaining examples also in the PR
> > > * b) keep only hello-world-yaml in the PR, but fix the description to not say "fixes #133", maybe instead say "partial fix for #133"
> > > 
> > > What are your thoughts about this?
> > 
> > 
> > I was testing only this hello-world-yaml example bcs , i believe other example are being tested via operrator-sdk
> 
> But seems Test's are not similar to what is mentioned in README.md

I added other two examples in e2e testing script .  We can not test `openshift-provision` example via Travis i guess.

scripts/e2e-test.sh
```diff
@@ -78,3 +78,144 @@ helm template deploy/helm/eunomia-operator/ \
   --set eunomia.operator.image.tag=dev \
   --set eunomia.operator.namespace=$TEST_NAMESPACE | kubectl delete -f -
 
+## Below block is to ensure all example are working
+helm template deploy/helm/eunomia-operator/ \
+  --set eunomia.operator.image.tag=dev \
+  --set eunomia.operator.image.pullPolicy=Never | kubectl apply -f -
+
+timeout=300
```
akavel: Please set timeout separately before each test of example, if possible, instead of one timeout for all tests of examples.

shaikmdrafi: Agreed it will save time if any test fails .
Done with changes .

scripts/e2e-test.sh
seanmalloy: Looks like this was fixed.

Issue #124, opened 2019-10-31
Use shellcheck in CI to help avoid bugs in shell scripts


akavel:
https://github.com/koalaman/shellcheck is a "linter" for bash scripts, helping avoid a boatload of  dangerous pitfalls of bash scripting (like a good spirit guide whispering life-saving warnings to your ear when navigating a minefield of bash syntax...)

PR #261, opened 2020-01-22
Added steps for shellcheck

rajneesh-ops:
**Description**

Ensure shellcheck is running to check a shell script quality

Fixes https://github.com/KohlsTechnology/eunomia/issues/124

**Type of change**

**Checklist**

- [ ] Documentation updated

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -52,7 +52,7 @@ test-dirty: generate
 	# TODO: also check that there are no untracked files, e.g. extra .go and .yaml ones
 
 .PHONY: test
-test: check-fmt lint vet test-unit test-e2e
+test: check-fmt lint vet shellcheck test-unit test-e2e
 
 .PHONY: test-e2e
 test-e2e:
@@ -80,6 +80,12 @@ lint:
 .PHONY: vet
 vet:
 	VET_INPUT="$(shell go list ./... | grep -v /vendor/)"; GO111MODULE=on go vet $$VET_INPUT
+
+# TODO: improve the command to also check scripts without .sh extension
+# get_helm.sh is ignored because it is getting download from internet
+.PHONY: shellcheck
+shellcheck:
+	for file in $(shell find . -not -path "./vendor/*" -not -path "./get_helm.sh" -name "*.sh") ; do shellcheck $$file ; done
 
 .PHONY: e2e-test-images
 e2e-test-images: build
--- a/scripts/build-images.sh
+++ b/scripts/build-images.sh
@@ -1,5 +1,7 @@
 #!/usr/bin/env bash
 
+# shellcheck disable=SC2002,SC2155
+
 # Copyright 2019 Kohl's Department Stores, Inc.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
--- a/template-processors/base/bin/processParameters.sh
+++ b/template-processors/base/bin/processParameters.sh
@@ -1,5 +1,7 @@
 #!/usr/bin/env bash
 
+# shellcheck disable=SC2001
+
 # Copyright 2019 Kohl's Department Stores, Inc.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
@@ -51,7 +53,8 @@ if [ -n "${FOLDERS}" ]; then
     YAML_FILES="$(find "${DIR}" -maxdepth 1 -name \*.json -o -name \*.yaml -o -name \*.yml)"
 
     # merge the files
-    if [ ! -z "${YAML_FILES}" ]; then
+    if [ "${YAML_FILES}" ]; then
+      # shellcheck disable=SC2086
       goyq merge -i -x "${VALUES_FILE}" ${YAML_FILES}
     fi
   done
--- a/template-processors/base/bin/processTemplates.sh
+++ b/template-processors/base/bin/processTemplates.sh
@@ -18,4 +18,5 @@ set -euxo pipefail
 
 echo Processing Templates
 
-cp -R $CLONED_TEMPLATE_GIT_DIR/* $MANIFEST_DIR/
+# shellcheck disable=SC2086
+cp -R "$CLONED_TEMPLATE_GIT_DIR/"* "$MANIFEST_DIR"/
--- a/template-processors/base/bin/resourceManager.sh
+++ b/template-processors/base/bin/resourceManager.sh
@@ -1,5 +1,7 @@
 #!/usr/bin/env bash
 
+# shellcheck disable=SC2002,SC2155
+
 # Copyright 2019 Kohl's Department Stores, Inc.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
@@ -21,6 +23,7 @@ TAG_APPLIED="gitopsconfig.eunomia.kohls.io/applied"
 
 # this is needed because we want the current namespace to be set as default if a namespace is not specified.
 function setContext {
+  # shellcheck disable=SC2154
   $kubectl config set-context current --namespace="$(cat /var/run/secrets/kubernetes.io/serviceaccount/namespace)"
   $kubectl config use-context current
 }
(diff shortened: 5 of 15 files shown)


.travis.yml
```diff
@@ -45,6 +45,7 @@ before_script:
 
 
 script:
+  - bash -c 'shopt -s globstar nullglob; shellcheck **/*.{sh,ksh,bash}'
```
seanmalloy: Can we add this to the `Makefile` instead? I'd like to be able to run `make shellcheck` from my laptop for local development. Also, the `vendor` directory should be ignored.

rajneesh-ops: Suggested changes are done.

rajneesh-ops:
I am fixing the shell script for shellcheck

template-processors/base/bin/processParameters.sh
```diff
@@ -51,8 +53,8 @@ if [ -n "${FOLDERS}" ]; then
     YAML_FILES="$(find "${DIR}" -maxdepth 1 -name \*.json -o -name \*.yaml -o -name \*.yml)"
 
     # merge the files
-    if [ ! -z "${YAML_FILES}" ]; then
-      goyq merge -i -x "${VALUES_FILE}" ${YAML_FILES}
+    if [ -z "${YAML_FILES}" ]; then
+      goyq merge -i -x "${VALUES_FILE}" "${YAML_FILES}"
```
akavel: on this line, we cannot use `"` around YAML_FILES, as it is important that we pass all values from YAML_FILES as separate arguments to goyq; please use a different approach that preserves this functionality

template-processors/base/bin/resourceManager.sh
```diff
@@ -1,5 +1,7 @@
 #!/usr/bin/env bash
 
+# shellcheck disable=SC2002,SC2044,SC2154
```
akavel: 1. for SC2044, please don't disable it globally for whole file, but instead do explicit disable on each line where it is reported
2. for SC2154, similar as above - this one is very often a bug, so we definitively don't want to disable this for whole file

template-processors/base/bin/resourceManager.sh
```diff
@@ -39,7 +41,7 @@ function kube {
 function addLabels {
   local owner="$1"
   local timestamp="$2"
-  local tmpdir="$(mktemp -d)"
+  tmpdir="$(mktemp -d)"
```
akavel: please don't remove `local`, see comment I made earlier above

template-processors/base/bin/resourceManager.sh
```diff
@@ -56,8 +58,8 @@ function deleteByOldLabels {
   local owner="$1"
   local timestamp="${2:-}"
   # NOTE: removing componentstatus because it shows up unintended in ownedKinds: https://github.com/kubernetes/kubectl/issues/151#issuecomment-562578617
-  local allKinds="$(kube api-resources --verbs=list -o name  | egrep -iv '^componentstatus(es)?$' | paste -sd, -)"
-  local ownedKinds="$(kube get "$allKinds" --ignore-not-found \
+  allKinds="$(kube api-resources --verbs=list -o name  | grep -ivE '^componentstatus(es)?$' | paste -sd, -)"
+  ownedKinds="$(kube get "$allKinds" --ignore-not-found \
```
akavel: please don't remove `local` in above lines, and similar for all cases below

template-processors/base/bin/wrapper.sh
```diff
@@ -1,5 +1,7 @@
 #!/usr/bin/env bash
 
+# shellcheck disable=SC1090
```
akavel: please don't disable this globally for whole script, but just for single lines where it happens

template-processors/openshift-provision/bin/processParameters.sh
```diff
@@ -19,5 +19,5 @@ set -euxo pipefail
 echo "Processing parameters in ${CLONED_PARAMETER_GIT_DIR}"
 
 ansible-playbook /files/processParameters.yml \
-    ${TEMPLATE_PROCESSOR_ARGS:-} \
+    "${TEMPLATE_PROCESSOR_ARGS:-}" \
```
akavel: change adding `"` here is not correct, we need contents of the env variable to be passed to command as separate arguments

template-processors/openshift-provision/bin/processTemplates.sh
```diff
@@ -19,6 +19,6 @@ set -euxo pipefail
 echo "Applying resources in ${CLONED_TEMPLATE_GIT_DIR}"
 
 ANSIBLE_ROLES_PATH=/files/roles ansible-playbook /files/processTemplates.yml \
-  ${TEMPLATE_PROCESSOR_ARGS:-} \
+  "${TEMPLATE_PROCESSOR_ARGS:-}" \
```
akavel: same as above, we cannot add `"` here

scripts/build-images.sh
```diff
@@ -1,5 +1,7 @@
 #!/usr/bin/env bash
 
+# shellcheck disable=SC2002
```
akavel: I would suggest we add SC2002 to `.shellcheckrc` file in eunomia instead of disabling it explicitly in scripts

seanmalloy:
@rajneesh-ops here are now some git merge conflicts with file `scripts/deploy-to-local.sh` due to some other recently merged pull requests.

rajneesh-ops:
@akavel , .shellcheckrc is supported in the 0.7 version of shellcheck. Right now Travis ci doesn't have this version I guess. For this, I have to add the installation step for shellcheck 0.7 version. Shall I do it or can we skip this for now? Also, I am not sure why build is failing in Travis, can you help me with this?

rajneesh-ops:
@akavel, Can you please review?

template-processors/base/bin/processTemplates.sh
```diff
@@ -18,4 +18,5 @@ set -euxo pipefail
 
 echo Processing Templates
 
+# shellcheck disable=SC2086
 cp -R $CLONED_TEMPLATE_GIT_DIR/* $MANIFEST_DIR/
```
akavel: Please change this line to:

    cp -R "$CLONED_TEMPLATE_GIT_DIR"/* "$MANIFEST_DIR/"

template-processors/base/bin/wrapper.sh
```diff
@@ -1,5 +1,8 @@
 #!/usr/bin/env bash
 
+# shellcheck source=$HOME/envs.sh
+# shellcheck disable=SC1091
```
akavel: AFAIU, above 2 lines should not be used together. Instead, please remove l.3, and move l.4 above l.28 - if this will work.

akavel:
> .shellcheckrc is supported in the 0.7 version of shellcheck. Right now Travis ci doesn't have this version I guess. For this, I have to add the installation step for shellcheck 0.7 version. Shall I do it or can we skip this for now? Also, I am not sure why build is failing in Travis, can you help me with this?

In such case I believe it's OK to not use .shellcheckrc for now, thanks.

template-processors/base/bin/processTemplates.sh
rajneesh-ops: This is breaking the changes of e2e tests. I can add this in comment

rajneesh-ops:
@akavel, I am done with the suggested changes, please review and let me know if there any changes required otherwise I will squash the commits

.travis.yml
```diff
@@ -45,6 +45,7 @@ before_script:
 
 
 script:
+  - make shellcheck
```
akavel: @seanmalloy What would you think, better to keep this here separately, or to include it as part of the general `make test` target, in Makefile l.55? https://github.com/KohlsTechnology/eunomia/blob/612a4769ae0dd8bd0ab15a936dd0fae49dfafd3a/Makefile#L55

Makefile
```diff
@@ -52,7 +52,7 @@ test-dirty: generate
 	# TODO: also check that there are no untracked files, e.g. extra .go and .yaml ones
 
 .PHONY: test
-test: check-fmt lint vet test-unit test-e2e
+test: check-fmt lint vet test-unit test-e2e shellcheck
```
akavel: [optional] I would suggest putting `shellcheck` between `vet` and `test-unit`, so that we don't have to wait for the very long `test-e2e` to finish to see any errors in shell scripts

PR #262, opened 2020-01-22
unit test for race condition fix

### LINKED ISSUE #179: Race condition between jobs started by eunomia

safirh:
**Description**

Unit tests for race condition fix [PR](https://github.com/KohlsTechnology/eunomia/pull/237)

Fixes #179 

**Type of change**

**Checklist**

- [ ] Documentation updated

### DIFF
--- a/pkg/controller/gitopsconfig/controller.go
+++ b/pkg/controller/gitopsconfig/controller.go
@@ -235,7 +235,7 @@ func (r *Reconciler) createJob(jobtype string, instance *gitopsv1alpha1.GitOpsCo
 		return reconcile.Result{}, xerrors.Errorf("unable to list the jobs in namespace %q: %w", instance.Namespace, err)
 	}
 	for _, j := range jobList.Items {
-		if isOwner(instance, &j) && j.Status.Active != 0 {
+		if isOwner(instance, &j) && (j.Status.Active != 0 || j.Status.StartTime.IsZero()) {
 			log.Info("Job is already running for this instance, postponing new job creation", "instance", instance.Name, "job", j.Name)
 			return reconcile.Result{
 				Requeue:      true,
--- a/pkg/controller/gitopsconfig/controller_test.go
+++ b/pkg/controller/gitopsconfig/controller_test.go
@@ -451,6 +451,19 @@ func TestCreatingDeleteJob(t *testing.T) {
 		log.Error(err, "Update CRD", "Failed Updating CRD type of GitOpsConfig")
 	}
 
+	// Fakeclient is not updating the job status , inorder to create the new job we are
+	// Updating the job status manually for the existing job created by Reconcile.
+	job = findRunningJob(cl)
+	job.Status.Active = 0
+	job.Status.Succeeded = 1
+	job.Status.Failed = 0
+	job.Status.StartTime = &deleteTime
+	// Update the job with the status
+	err = cl.Update(context.TODO(), &job)
+	if err != nil {
+		log.Error(err, "Update job", "Failed to Updating the job status")
+	}
+
 	// There shouldn't be a delete job at this point, the reconciler should create one
 	r.Reconcile(req)
 
@@ -559,6 +572,92 @@ func findDeleteJob(cl client.Client) batch.Job {
 		if job.GetLabels()["action"] == "delete" {
 			return job
 		}
+	}
+	return batch.Job{}
+}
+
+func TestCreateJob(t *testing.T) {
+	// This flag is needed to let the reconciler know that the CRD has been initialized
+	gitops.Annotations = map[string]string{"gitopsconfig.eunomia.kohls.io/initialized": "true"}
+	// Set trigger type to Change
+	gitops.Spec.Triggers = []gitopsv1alpha1.GitOpsTrigger{
+		{
+			Type: "Change",
+		},
+	}
+	// Objects to track in the fake client.
+	objs := []runtime.Object{
+		gitops,
+	}
+	// Register operator types with the runtime scheme.
+	s := scheme.Scheme
+	s.AddKnownTypes(gitopsv1alpha1.SchemeGroupVersion, gitops)
+	// Initialize fake client
+	cl := fake.NewFakeClient(objs...)
+	r := &Reconciler{client: cl, scheme: s}
+
+	nsn := types.NamespacedName{
+		Name:      name,
+		Namespace: namespace,
+	}
+
+	req := reconcile.Request{
+		NamespacedName: nsn,
+	}
+
+	r.Reconcile(req)
+
+	// Fakeclient is not updating the job status , inorder to test race condition between the jobs we are
+	// Updating the job status manually for the existing job created by Reconcile.
+	startTime := metav1.Now()
+	job := findRunningJob(cl)
+	job.Status.Active = 1
+	job.Status.Succeeded = 0
+	job.Status.Failed = 0
+	job.Status.StartTime = &startTime
+
+	err := cl.Update(context.TODO(), &job)
+	if err != nil {
+		log.Error(err, "Update job", "Failed to Updating the job status")
+	}
+	r.Reconcile(req)
+	jobCount, err := findJobList(cl)
+	if err != nil {
+		log.Error(err, "Job list", "Failed to fetch job list")
+	}
+	if jobCount > 1 {
+		t.Error("Job was not postponed")
+	}
+}
+
+func findJobList(cl client.Client) (int, error) {
+	// At times other jobs can exist
+	jobList := &batchv1.JobList{}
+	// Looking up all jobs
+	err := cl.List(context.TODO(), &client.ListOptions{
+		Namespace: namespace,
+	}, jobList)
+	if err != nil {
+		log.Error(err, "unable to list the running jobs")
+		return 0, err
+	}
+	return len(jobList.Items), nil
+}
+
+func findRunningJob(cl client.Client) batch.Job {
+	// At times other jobs can exist
+	jobList := &batchv1.JobList{}
+	// Looking up all jobs
+	err := cl.List(context.TODO(), &client.ListOptions{
+		Namespace: namespace,
+	}, jobList)
+	if err != nil {
+		log.Error(err, "unable to list jobs")
+		return batch.Job{}
+	}
+	// Returning the jobs
+	if len(jobList.Items) > 0 {
+		return jobList.Items[0]
 	}
 	return batch.Job{}
 }


pkg/controller/gitopsconfig/controller.go
```diff
@@ -235,7 +235,7 @@ func (r *Reconciler) createJob(jobtype string, instance *gitopsv1alpha1.GitOpsCo
 		return reconcile.Result{}, xerrors.Errorf("unable to list the jobs in namespace %q: %w", instance.Namespace, err)
 	}
 	for _, j := range jobList.Items {
-		if isOwner(instance, &j) && j.Status.Active != 0 {
+		if isOwner(instance, &j) && (j.Status.Active != 0 || j.Status.StartTime.IsZero()) && instance.DeletionTimestamp.IsZero() {
```
akavel: Based on our talk on chat, I now understand that the check of `DeletionTimestamp.IsZero` is a workaround to make TestCreatingDeleteJob pass, and I believe I understand more about the context why you did it this way. Thanks to your explanations, I also now finally understood the meaning of the idea you said before, that I didn't understand at that time — to modify Job.Status in the "fake client" used in TestCreatingDeleteJob. Based on this, I believe this idea you had was the correct solution here, so I would like to ask you to change it now, to:

 1. remove the `IsZero` check from this line
 2. in TestCreatingDeleteJob, add code that will simulate a change of the action=create Job Status to "completed", to allow Reconcile to make a new action=delete job

pkg/controller/gitopsconfig/controller_test.go
```diff
@@ -562,3 +564,66 @@ func findDeleteJob(cl client.Client) batch.Job {
...
+	req := reconcile.Request{
+		NamespacedName: nsn,
+	}
+
+	numberOfJobs := 2
+	var wg sync.WaitGroup
+	for i := 1; i <= numberOfJobs; i++ {
+		wg.Add(1)
+		// in order to replicate reconciler being called for creating an another job.
+		go func() {
```
akavel: Instead of running goroutines, can we just call Reconcile two times, one after another? or are there reasons why we cannot or should not do this?

pkg/controller/gitopsconfig/controller_test.go
```diff
@@ -562,3 +564,66 @@ func findDeleteJob(cl client.Client) batch.Job {
...
+			r.Reconcile(req)
```
akavel: Instead of running Reconcile, could we just call `r.createJob`? or are there reasons why we cannot or should not do this?

pkg/controller/gitopsconfig/controller_test.go
```diff
@@ -562,3 +564,66 @@ func findDeleteJob(cl client.Client) batch.Job {
...
+			wg.Done()
+		}()
+		time.Sleep(3 * time.Second)
```
akavel: Why we need to run Sleep here? In general, it is much better if all tests can be made to run as quickly as possible, to avoid wasting everybody's time; adding Sleep makes tests run much longer, so it is usually better if tests can be modified in some way to still work correctly, but not using Sleep or other time delays, or at least to make them as short as possible if they are absolutely necessary

pkg/controller/gitopsconfig/controller_test.go
```diff
@@ -562,3 +564,66 @@ func findDeleteJob(cl client.Client) batch.Job {
...
+	}
+	wg.Wait()
+
+	job, _ := findJobList(cl)
```
akavel: please don't ignore errors from functions, but instead, make sure test fails when there is error from function, for example:

```
if err != nil {
    t.Error(err)
}
```

pkg/controller/gitopsconfig/controller_test.go
```diff
@@ -453,6 +451,19 @@ func TestCreatingDeleteJob(t *testing.T) {
 		log.Error(err, "Update CRD", "Failed Updating CRD type of GitOpsConfig")
 	}
 
+	// fakeclient is not updating the job status , inorder to create the job with different action
+	//updating the job status manually
```
akavel: Please also add information here explaining that this job we are updating exists because it was created by  Reconcile.

pkg/controller/gitopsconfig/controller_test.go
safirh: Changes done

pkg/controller/gitopsconfig/controller_test.go
safirh: Changes done

pkg/controller/gitopsconfig/controller_test.go
safirh: I followed this pattern as we have used reconcile in all other test cases as a standard rather than using a createJob.

pkg/controller/gitopsconfig/controller_test.go
safirh: changes done

pkg/controller/gitopsconfig/controller.go
safirh: Changes done

pkg/controller/gitopsconfig/controller_test.go
```diff
@@ -595,21 +606,28 @@ func TestCreateJob(t *testing.T) {
...
-	job, _ := findJobList(cl)
+	err := cl.Update(context.TODO(), &job)
+	if err != nil {
+		log.Error(err, "Update job", "Failed to Updating the job status")
+	}
+	r.Reconcile(req)
+	jobCount, err := findJobList(cl)
 
-	if job == numberOfJobs {
+	if err != nil {
```
akavel: Please remove empty line above this check; in Go it is best style when `if err != nil {` check line is as close as possible to the line where the `err` it is checking was set - this is a very important check, and keeping it as close as possible makes it easiest to immediately see that it was not forgotten.

pkg/controller/gitopsconfig/controller_test.go
```diff
@@ -595,21 +606,28 @@ func TestCreateJob(t *testing.T) {
...
+		log.Error(err, "Job list", "Failed to fetch job list")
+	}
+	if jobCount == numberOfJobs {
```
akavel: I believe it would be better to remove `numberOfJobs` variable, and just check here `if jobCount > 1`. If I understand correctly, in this test we want to verify that no new jobs were created, regardless of exact number - too many jobs (more than 2) would be also incorrect.

pkg/controller/gitopsconfig/controller_test.go
```diff
@@ -627,3 +645,21 @@ func findJobList(cl client.Client) (int, error) {
...
+	jobList := &batchv1.JobList{}
+	// Looking up all jobs
+	err := cl.List(context.TODO(), &client.ListOptions{
+		Namespace: namespace,
+	}, jobList)
+	if err != nil {
+		log.Error(err, "unable to list jobs")
+		return batch.Job{}
+	}
+	// Return the first instance that is a delete job
```
akavel: This comment says about a delete job, are you sure this is correct?

pkg/controller/gitopsconfig/controller_test.go
```diff
@@ -562,3 +575,91 @@ func findDeleteJob(cl client.Client) batch.Job {
...
+	for _, job := range jobList.Items {
+		return job
+	}
```
akavel: please change above block to:

```
if len(jobList.Items) > 0 {
    return jobList.Items[0]
}
```

Using `for` here is technically correct, but using `if` better conveys the idea that we want to return at most 1 result; using `for` suggests we want to do something more than once with elements of jobList.Items, which makes it misleading.

safirh: changes done

pkg/controller/gitopsconfig/controller_test.go
safirh: fixed the comment

pkg/controller/gitopsconfig/controller_test.go
safirh: changes done

pkg/controller/gitopsconfig/controller_test.go
safirh: removed empty line

pkg/controller/gitopsconfig/controller_test.go
safirh: changes done

pkg/controller/gitopsconfig/controller.go
```diff
@@ -235,7 +235,7 @@ func (r *Reconciler) createJob(jobtype string, instance *gitopsv1alpha1.GitOpsCo
 		return reconcile.Result{}, xerrors.Errorf("unable to list the jobs in namespace %q: %w", instance.Namespace, err)
 	}
 	for _, j := range jobList.Items {
-		if isOwner(instance, &j) && (j.Status.Active != 0 || j.Status.StartTime.IsZero()) && instance.DeletionTimestamp.IsZero() {
+		if isOwner(instance, &j) && j.Status.Active != 0 {
```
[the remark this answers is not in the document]
safirh: reverted back

pkg/controller/gitopsconfig/controller_test.go
```diff
@@ -451,8 +451,8 @@ func TestCreatingDeleteJob(t *testing.T) {
 		log.Error(err, "Update CRD", "Failed Updating CRD type of GitOpsConfig")
 	}
 
-	// fakeclient is not updating the job status , inorder to create the job with different action
-	//updating the job status manually
+    // fakeclient is not updating the job status , inorder to create the new job we are
```
akavel: fix gofmt to fix Travis fail

safirh: gofmt fixed

PR #263, opened 2020-01-22
Set eunomia-operator user to non-root

### LINKED ISSUE #171: Let the eunomia-operator pod user be non-root regardless of environment

When I exec to eunomia-operator pod in **openshift**, and I issue `whoami` command, I get
```
sh-4.4$ whoami
whoami: cannot find name for user ID 1000500000
```
which is ok because it looks like a generic UID.

On the other hand when I exec to the same pod in **k8s**, and I issue the exact same command, I get
```
sh-4.4$ whoami
root
```

User could be set explicitly in Dockerfile as described in [Dockerfile best practices](https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#user) so the user would be non-root regardless of environment.

n/a

eunomia version: v0.0.4

**Additional context**

karolkrawczyk:
**Description**

In order to follow the "least privilege" security best practice,
set eunomia-operator user to non-root.

Fixes #171 

**Type of change**

**Checklist**

- [not needed] Unit tests and e2e tests updated
- [not needed] Documentation updated

### DIFF
--- a/build/Dockerfile
+++ b/build/Dockerfile
@@ -22,4 +22,8 @@ COPY job-templates/* /default-job-templates/
 # install operator binary
 COPY _output/bin/eunomia ${OPERATOR}
 
+RUN groupadd -r eunomia && useradd --no-log-init -r -g eunomia eunomia
+
+USER eunomia
+
 ENTRYPOINT ["/usr/local/bin/eunomia-operator"]


build/Dockerfile
```diff
@@ -23,3 +23,5 @@ COPY job-templates/* /default-job-templates/
 COPY _output/bin/eunomia ${OPERATOR}
 
 ENTRYPOINT ["/usr/local/bin/eunomia-operator"]
+
+USER 12345
```
Smiley73: Does this work without the UID even existing?
On openshift it gets overwritten by the platform, but it might now work an a plain K8s (like GKE)

karolkrawczyk: Yes it does. As per docker documentation it is possible to just provide UID: https://docs.docker.com/engine/reference/builder/#user. Also I tested it locally on minikube and I haven't encountered any problems.

seanmalloy: We also need to make sure that the group is not root. So how about this instead ...

```
RUN groupadd -r eunomia && useradd --no-log-init -r -g eunomia eunomia
USER eunomia
```

karolkrawczyk: Sure, done.

build/Dockerfile
```diff
@@ -22,4 +22,8 @@ COPY job-templates/* /default-job-templates/
 # install operator binary
 COPY _output/bin/eunomia ${OPERATOR}
 
+RUN groupadd -r eunomia && useradd --no-log-init -r -g eunomia eunomia
+
 ENTRYPOINT ["/usr/local/bin/eunomia-operator"]
+
+USER eunomia
```
seanmalloy: I think `USER` needs to be before `ENTRYPOINT`. I think after that it should be good to merge.

From ... https://docs.docker.com/engine/reference/builder/#user ...

> The USER instruction sets the user name (or UID) and optionally the user group (or GID) to use when running the image and for any RUN, CMD and ENTRYPOINT instructions **that follow it** in the Dockerfile.

karolkrawczyk: Oooops, I think you might be right. Thank you for noticing.

PR #265, opened 2020-01-22
Run each e2e test function in a separate namespace

### LINKED ISSUE #211: Each e2e test should run in separate namespace

Currently, all e2e tests run in a single namespace `test-eunomia-operator`. The problem with this is such, that some remains from a previous test, that were not yet fully cleaned up when the test ends, can leak into a subsequent test, changing its output (even changing "fail" to "success" or opposite). Exactly something like this happened to me when writing tests, and was fairly tricky to debug, and in the end forced me to do a workaround.

I'd like if each e2e test was running in a separate namespace.

 - I also considered adding a wait for "no pods in current namespace" at the beginning of each test. This is however still fragile, and prone to missing other stuff. Creating namespaces from scratch should provide much better and more robust isolation. 
 - Currently I used an alternative workaround of templates with pods named differently for each test, so that I can detect pods by different name in each test. This is however introducing unnecessary repetition and trash in the repo as a workaround.

eunomia version: v0.0.5

**Additional context**

As an extra benefit, using separate namespaces could potentially make it super easy to [parallelize the tests](https://golang.org/pkg/testing/#T.Parallel), so that the full test suite could end faster. This is especially important with tests which are long because of some intrinsically long kubernetes behaviors, such as exponential backoff for failing pods in Jobs.

sbar95:
**Description**

Run each test function in separate namespace.

Add SetupRbacInNamespace function to make it possible to deploy a
service account to on-the-fly generated namespace (during tests).

In each test deploy eunomia-operator service account bound to
eunomia-operator role. Also remove underscores from test function names
to make per-namespace tests possible because test namespaces created
dynamically by operator-sdk framework take function name as part of a
namespace (and such cannot have an underscore).

Make it possible to use TEMPLATE_PROCESSOR_ARGS env variable in helm
template processor.

Fixes #211 

**Type of change**

**Checklist**

- [ ] Documentation updated (no need for this PR)

### DIFF
--- a/deploy/helm/eunomia-operator/templates/deployment.yaml
+++ b/deploy/helm/eunomia-operator/templates/deployment.yaml
@@ -1,5 +1,5 @@
 {{- with .Values.eunomia.operator }}
-{{- if .deployment.enabled -}}
+{{- if and .deployment.enabled (not .deployment.nsRbacOnly) -}}
 apiVersion: apps/v1
 kind: Deployment
 metadata:
--- a/scripts/e2e-test.sh
+++ b/scripts/e2e-test.sh
@@ -34,8 +34,8 @@ fi
 echo "EUNOMIA_URI=${EUNOMIA_URI:-}"
 echo "EUNOMIA_REF=${EUNOMIA_REF:-}"
 
-# Ensure minikube is running
-#minikube start
+# Check if minikube is running
+minikube status || { echo "Minikube is not running, aborting tests"; exit 1; }
 
 # Ensure clean workspace
 if [[ $(kubectl get namespace $TEST_NAMESPACE) ]]; then
@@ -47,41 +47,29 @@ fi
 eval $(minikube docker-env)
 GOOS=linux make e2e-test-images
 
-helm template deploy/helm/eunomia-operator/ \
-  --set eunomia.operator.deployment.enabled= \
-  --set eunomia.operator.namespace=$TEST_NAMESPACE | kubectl apply -f -
-
-operator-sdk test local ./test/e2e --namespace "$TEST_NAMESPACE" --up-local --no-setup --verbose --go-test-flags "-tags e2e -timeout 20m"
-
-helm template deploy/helm/eunomia-operator/ \
-  --set eunomia.operator.deployment.enabled= \
-  --set eunomia.operator.namespace=$TEST_NAMESPACE | kubectl delete -f -
-
+# Eunomia setup
 helm template deploy/helm/eunomia-operator/ \
   --set eunomia.operator.image.tag=dev \
   --set eunomia.operator.image.pullPolicy=Never \
   --set eunomia.operator.namespace=$TEST_NAMESPACE | kubectl apply -f -
 
+# Deployment test
 kubectl wait --for=condition=available --timeout=30s deployment/eunomia-operator -n $TEST_NAMESPACE
-
 podname=$(kubectl get pods  -o jsonpath='{range .items[*]}{.metadata.name}{"\n"}{end}' -n $TEST_NAMESPACE)
-
 if kubectl exec "${podname}" date -n $TEST_NAMESPACE
 then
-  echo "Pod is Healthy"
+  echo "Eunomia deployment successful"
 else
-  echo "Pod is not Healthy"
+  echo "Eunomia deployment failed"
   exit 1
 fi
 
-helm template deploy/helm/eunomia-operator/ \
-  --set eunomia.operator.image.tag=dev \
-  --set eunomia.operator.namespace=$TEST_NAMESPACE | kubectl delete -f -
-
-## Below block is to ensure all example are working
-helm template deploy/helm/eunomia-operator/ \
-  --set eunomia.operator.image.tag=dev \
-  --set eunomia.operator.image.pullPolicy=Never | kubectl apply -f -
+# End-to-end tests
+operator-sdk test local ./test/e2e \
+  --namespaced-manifest /dev/null \
+  --global-manifest /dev/null \
+  --verbose \
+  --go-test-flags "-tags e2e -timeout 20m"
 
 ## Testing hello-world-yaml example
 # Create new namespace
@@ -226,3 +214,8 @@ hello_world_hierarchy_cr1
 # Delete namespaces after Testing hello-world-hierarchy example
 kubectl delete namespace eunomia-hello-world-demo eunomia-hello-world-demo-hierarchy
 
+# Eunomia teardown
+helm template deploy/helm/eunomia-operator/ \
+  --set eunomia.operator.image.tag=dev \
+  --set eunomia.operator.image.pullPolicy=Never \
+  --set eunomia.operator.namespace=$TEST_NAMESPACE | kubectl delete -f -
--- a/template-processors/helm/bin/processTemplates.sh
+++ b/template-processors/helm/bin/processTemplates.sh
@@ -24,4 +24,9 @@ helm init --client-only
 helm repo update "${CLONED_TEMPLATE_GIT_DIR}"
 
 echo "Generating manifest files"
-helm template -f /tmp/eunomia_values_processed.yaml --output-dir "${MANIFEST_DIR}" --namespace "${NAMESPACE}" "${CLONED_TEMPLATE_GIT_DIR}"
+helm template \
+  -f /tmp/eunomia_values_processed.yaml \
+  ${TEMPLATE_PROCESSOR_ARGS:-} \
+  --output-dir "${MANIFEST_DIR}" \
+  --namespace "${NAMESPACE}" \
+  "${CLONED_TEMPLATE_GIT_DIR}"
--- a/test/e2e/util.go
+++ b/test/e2e/util.go
@@ -24,6 +24,9 @@ import (
 	"encoding/json"
 	"fmt"
 	"io"
+	"os/exec"
+	"path/filepath"
+	"runtime"
 	"strings"
 	"testing"
 	"time"
@@ -189,4 +192,45 @@ func DumpJobsLogsOnError(t *testing.T, f *framework.Framework, namespace string)
 func debugJSON(v interface{}) string {
 	raw, _ := json.MarshalIndent(v, "", "  ")
 	return string(raw)
+}
+
+// SetupRbacInNamespace deploys appropriate role and its binding to service
+// account in given namespace for the test to be able to run seccessfuly.
+//
+// This function is needed because the function which normally handles
+// per-namespace recource creation in operator-sdk framework
+// https://github.com/operator-framework/operator-sdk/blob/v0.8.1/pkg/test/resource_creator.go#L108
+// does not do the job in eunomia's case. It returns a call to a function
+// CreateFromYAML (defined just above it) which uses SetNamespace method which
+// sets a namespace in the yaml provided with "operator-sdk test local
+// --namespaced-manifest" flag. However, this function
+// https://github.com/kubernetes/apimachinery/blob/master/pkg/apis/meta/v1/unstructured/unstructured.go#L237-L243
+// sets the namespace only in the metadata section of the resource. It is not
+// enough in eunomia's case because the namespace needs to be also set in
+// RoleBinding's subjects section:
+// https://github.com/KohlsTechnology/eunomia/blob/v0.1.1/deploy/helm/eunomia-operator/templates/service-account-operator.yaml#L17
+func SetupRbacInNamespace(namespace string) error {
+	// get eunomia root dir
+	_, utilTestFilename, _, _ := runtime.Caller(0)
+	eunomiaRoot := filepath.Join(filepath.Dir(utilTestFilename), "../..")
+
+	// create kubectl input
+	out, err := exec.Command(
+		"helm", "template",
+		filepath.Join(eunomiaRoot, "deploy/helm/eunomia-operator/"),
+		"--set", "eunomia.operator.namespace="+namespace,
+		"--set", "eunomia.operator.deployment.nsRbacOnly=true",
+	).Output()
+	if err != nil {
+		return xerrors.Errorf("Failed to generate manifest files: %w", err)
+	}
+
+	// create role, role binding and service account
+	cmd := exec.Command("kubectl", "apply", "-f", "-")
+	cmd.Stdin = bytes.NewReader(out)
+	if err := cmd.Run(); err != nil {
+		return xerrors.Errorf("Failed to create RBAC in %s namespace: %w", namespace, err)
+	}
+
+	return nil
 }
--- a/deploy/helm/eunomia-operator/templates/clusterrole-gitopsconfig-admin.yaml
+++ b/deploy/helm/eunomia-operator/templates/clusterrole-gitopsconfig-admin.yaml
@@ -1,3 +1,4 @@
+{{- if not .Values.eunomia.operator.deployment.nsRbacOnly -}}
 apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRole
 metadata:
@@ -13,3 +14,4 @@ rules:
   - 'gitopsconfigs'
   verbs:
   - '*'
+{{- end }}
--- a/deploy/helm/eunomia-operator/templates/clusterrole-gitopsconfig-viewer.yaml
+++ b/deploy/helm/eunomia-operator/templates/clusterrole-gitopsconfig-viewer.yaml
@@ -1,3 +1,4 @@
+{{- if not .Values.eunomia.operator.deployment.nsRbacOnly -}}
 apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRole
 metadata:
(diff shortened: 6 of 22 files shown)


template-processors/helm/bin/processTemplates.sh
```diff
@@ -24,4 +24,5 @@ helm init --client-only
 helm repo update "${CLONED_TEMPLATE_GIT_DIR}"
 
 echo "Generating manifest files"
-helm template -f /tmp/eunomia_values_processed.yaml --output-dir "${MANIFEST_DIR}" --namespace "${NAMESPACE}" "${CLONED_TEMPLATE_GIT_DIR}"
+helm template -f /tmp/eunomia_values_processed.yaml ${TEMPLATE_PROCESSOR_ARGS:-} \
+  --output-dir "${MANIFEST_DIR}" --namespace "${NAMESPACE}" "${CLONED_TEMPLATE_GIT_DIR}"
```
akavel: 1. In the commit message that introduces the changes here, could you please add a reference to the issue where TEMPLATE_PROCESSOR_ARGS were originally introduced? This could make it easier to understand the bigger context of this change 
2. [optional] What would you think about splitting this command such that each flag+parameter is on a separate line?

deploy/test/eunomia-rbac/templates/role-operator.yaml
```diff
@@ -0,0 +1,55 @@
...
+# Metrics  
+- apiGroups:
+  - apps
+  resources:
+  - deployments/finalizers
+  resourceNames:
+  - eunomia-operator
+  verbs:
+  - "update"
+{{- end }}
```
akavel: How does this file differ from the already existing file `deploy/helm/eunomia-operator/templates/role-operator.yaml`? why can't we reuse that existing one and we have to create a new very similar (or is it identical?) file?

test/e2e/util.go
```diff
@@ -1,4 +1,4 @@
-// +build e2e
+// +build e2e xd
```
akavel: What's the `xd` here for?

scripts/e2e-test.sh
```diff
@@ -51,7 +51,14 @@ helm template deploy/helm/eunomia-operator/ \
...
 
-operator-sdk test local ./test/e2e --namespace "$TEST_NAMESPACE" --up-local --no-setup --verbose --go-test-flags "-tags e2e -timeout 20m"
+go test ./test/e2e/... \
+  -namespacedMan /dev/null \
+  -globalMan /dev/null \
+  -root "$EUNOMIA_PATH" \
+  -test.parallel=1 \
+  -localOperator \
+  -tags e2e \
+  -v
```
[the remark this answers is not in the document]
seanmalloy: I'm curious on this one too. Can we ask the `operator-sdk` maintainers if there is a way to make this work? Maybe we can open an issue or post in their slack channel?

deploy/test/eunomia-rbac/templates/role-operator.yaml
seanmalloy: I don't think we should have special RBAC that is only deployed when running the tests. We should just you our standard RBAC configs when running the tests.

scripts/e2e-test.sh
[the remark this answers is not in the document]
sbar95: I cannot run `operator-sdk test local`'s `--up-local` flag without the `--namespace` flag:
https://github.com/operator-framework/operator-sdk/blob/v0.8.1/cmd/operator-sdk/test/local.go#L116-L117 (it's also the case in the latest operator-sdk).
<details><summary><code>operator-sdk test local --help</code></summary><br><pre>
$ operator-sdk test local --help
Run End-To-End tests locally

Usage:
  operator-sdk test local <path to tests directory> [flags]

Flags:
      --debug                        Enable debug-level logging
      --global-manifest string       Path to manifest for Global resources (e.g. CRD manifests)
      --go-test-flags string         Additional flags to pass to go test
  -h, --help                         help for local
      --image string                 Use a different operator image from the one specified in the namespaced manifest
      --kubeconfig string            Kubeconfig path
      --molecule-test-flags string   Additional flags to pass to molecule test
      --namespace string             If non-empty, single namespace to run tests in
      --namespaced-manifest string   Path to manifest for per-test, namespaced resources (e.g. RBAC and Operator manifest)
      --no-setup                     Disable test resource creation
      --up-local                     Enable running operator locally with go run instead of as an image in the cluster

Global Flags:
      --verbose   Enable verbose logging
</pre></details>
However, it's still possible to run the tests successfully by issuing `operator-sdk test local ./test/e2e --no-setup --verbose --go-test-flags "-tags e2e -timeout 20m"` (deploying eunomia to the cluster beforehand) **but** then all the tests are being run in the `default` namespace (and that's because of [GetNamespace() function](https://github.com/operator-framework/operator-sdk/blob/v0.8.1/pkg/test/resource_creator.go#L36-L39) - `*singleNamespace` is in turn enforced by [`--no-setup` flag](https://github.com/operator-framework/operator-sdk/blob/v0.8.1/cmd/operator-sdk/test/local.go#L205-L207)).

**Finally**
Well, in course of writing this comment I actually found the right `oprator-sdk`'s flag combination to run the tests successfully in separate namespaces (no need to resort to bare `go test ...`), but eunomia has to be deployed to the cluster beforehand (I cannot use `--up-local`) but I think this is fine, because apart from e2e tests we now have a deployment test at the same time (so we don't need a [separate test](https://github.com/KohlsTechnology/eunomia/blob/master/scripts/e2e-test.sh#L60-L80) at the end of the `e2e-test.sh` script - and if the deployment fails, e2e tests just won't run):
```
operator-sdk test local ./test/e2e \
--namespaced-manifest /dev/null \
--global-manifest /dev/null \
--verbose \
--go-test-flags "-tags e2e -timeout 20m"
```
Soooo... Thank you for the comments, they helped me a lot! 👍

test/e2e/util.go
```diff
@@ -189,3 +194,43 @@ func debugJSON(v interface{}) string {
...
+		return fmt.Errorf("Failed to create temporary file")
+	}
+	defer os.Remove(tmpFile.Name())
+
+	// create kubectl input
+	out, err := exec.Command(
+		"helm", "template",
+		filepath.Join(eunomiaRoot, "deploy/test/eunomia-rbac/"),
+		"--set", "eunomia.namespace="+namespace,
+	).Output()
```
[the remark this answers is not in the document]
sbar95: There is the [InitializeClusterResources() function](https://github.com/operator-framework/operator-sdk/blob/v0.8.1/pkg/test/resource_creator.go#L108) using namespacedMan parameter I didn't see earlier - I'll try to come up with something using it (but I will also need a namespaced manifest dir - maybe I'll be generating it dynamically somehow).

deploy/test/eunomia-rbac/templates/role-operator.yaml
sbar95: This is the standard test RBAC, but per namespace. Until now it was being deployed each time to the [sole test namespace `test-eunomia-operator`](https://github.com/KohlsTechnology/eunomia/blob/master/scripts/e2e-test.sh#L52) with the use of [`helm template`](https://github.com/KohlsTechnology/eunomia/blob/master/scripts/e2e-test.sh#L50-L52):
- https://github.com/KohlsTechnology/eunomia/blob/master/deploy/helm/eunomia-operator/templates/role-operator.yaml
- https://github.com/KohlsTechnology/eunomia/blob/master/deploy/helm/eunomia-operator/templates/service-account-operator.yaml

We had these in only one test namespace (because all tests were run in only one). Now with each test function being run in a separate test namespace I need the Role, ServiceAccount, and RoleBinding in each namespace for the tests not to raise access errors.

test/e2e/util.go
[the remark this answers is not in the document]
sbar95: The function I linked to isn't enough unfortunately. I can create a single manifest file like this one:
<details><summary><code>my_manifest.yml</code></summary><br><pre>
---
# Source: eunomia-rbac/templates/service-account-operator.yaml
# service account for the operator
apiVersion: v1
kind: ServiceAccount
metadata:
  name: eunomia-operator
  namespace: eunomia-operator
---
# Source: eunomia-rbac/templates/role-operator.yaml
kind: Role
apiVersion: rbac.authorization.k8s.io/v1
metadata:
  name: eunomia-operator
  namespace: "eunomia-operator"
rules:
# leader election   
- apiGroups:
  - ""
  resources:
  - configmaps
  - pods
  verbs:
  - "*"  
#Metrics  
- apiGroups:
  - ""
  resources:
  - services
  verbs:
  - "*"
#Metrics  
- apiGroups:
  - "apps"
  resources:
  - replicasets
  - deployments
  verbs: 
  - "*"
# In Kubernetes <1.16, "deployments" also require API group "extensions" (see: https://stackoverflow.com/a/57854939)
- apiGroups:
  - "extensions"
  resources:
  - deployments
  verbs:
  - "*"
# Metrics  
- apiGroups:
  - monitoring.coreos.com
  resources:
  - servicemonitors
  verbs:
  - "get"
  - "create"
# Metrics  
- apiGroups:
  - apps
  resources:
  - deployments/finalizers
  resourceNames:
  - eunomia-operator
  verbs:
  - "update"
---
# Source: eunomia-rbac/templates/service-account-operator.yaml
kind: RoleBinding
apiVersion: rbac.authorization.k8s.io/v1
metadata:
  name: eunomia-operator
  namespace: eunomia-operator
subjects: 
- kind: ServiceAccount
  name: eunomia-operator
  namespace: eunomia-operator
roleRef:
  kind: Role
  name: eunomia-operator
  apiGroup: rbac.authorization.k8s.io
</pre></details>

and pass it to operator-sdk: (`operator-sdk test local ./test/e2e --namespaced-manifest /path/to/my_manifest.yml`.

But this is what happens:
In each test function I invoke:
- [`InitializeClusterResources`](https://github.com/operator-framework/operator-sdk/blob/v0.8.1/pkg/test/resource_creator.go#L108),
- which returns [`ctx.createFromYAML`](https://github.com/operator-framework/operator-sdk/blob/v0.8.1/pkg/test/resource_creator.go#L55),
- this function [sets the current namespace](https://github.com/operator-framework/operator-sdk/blob/v0.8.1/pkg/test/resource_creator.go#L72) in the resources provided by my `my_manifest.yml`,
- but unfortunately only in the [metadata section](https://github.com/kubernetes/apimachinery/blob/master/pkg/apis/meta/v1/unstructured/unstructured.go#L237-L243).

This is a problem for me because I need the namespace to be also set in `my_manifest.yml` in RoleBinding's subjects subsection. It doesn't happen and the namespace there stays set to the initial value (it isn't being changed per test to the dynamically created namespace as in the metadata section) so I get access errors.

deploy/test/eunomia-rbac/templates/role-operator.yaml
sbar95: @akavel I'll try to reuse the already existing files

sbar95: They are now being reused: 22073da.
Later, I'll squash it and edit this message too.

test/e2e/util.go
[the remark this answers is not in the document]
sbar95: After additional research I found out that if you don't pass `--namespaced-manifest` flag with content to the `operator-sdk test local` command (and at the same time you don't provide `--no-setup` flag) the program looks for [some default files](https://github.com/operator-framework/operator-sdk/blob/v0.8.1/cmd/operator-sdk/test/local.go#L122) which we do not have in `deploy/` directory (maybe they were created by the `operator-sdk` initially but then removed at some point in time because we started using helm?).

sbar95:
@seanmalloy @akavel 
Please, re-review. I pushed some additional commits addressing the comments. I also put some answers.
Please, if it's acceptable, don't merge for now - I'm going to squash my commits then and remove the WIP label.

test/e2e/util.go
```diff
@@ -189,3 +194,57 @@ func debugJSON(v interface{}) string {
...
+
+	// create kubectl input
+	out, err := exec.Command(
+		"helm", "template",
+		filepath.Join(eunomiaRoot, "deploy/helm/eunomia-operator/"),
+		"--set", "eunomia.operator.namespace="+namespace,
+		"--set", "eunomia.operator.deployment.nsRbacOnly=true",
+	).Output()
+	if err != nil {
+		return fmt.Errorf("Failed to generate manifest files")
```
akavel: Please don't lose the value present in `err` - please use xerrors.Errorf + %w format:

    return xerrors.Errorf("Failed to generate manifest files: %w", err)

— please also fix **in all other places in this func**

test/e2e/util.go
```diff
@@ -189,3 +194,57 @@ func debugJSON(v interface{}) string {
...
+// enough in eunomia's case because the namespace needs to be also set in
+// RoleBinding's subjects section:
+// https://github.com/KohlsTechnology/eunomia/blob/v0.1.1/deploy/helm/eunomia-operator/templates/service-account-operator.yaml#L17
+func SetupRbacInNamespace(namespace string) error {
+	// get eunomia root dir
+	_, utilTestFilename, _, _ := runtime.Caller(0)
+	eunomiaRoot := filepath.Join(filepath.Dir(utilTestFilename), "../..")
+
+	// create temporary file to store kubectl input in
+	tmpFile, err := ioutil.TempFile(os.TempDir(), "sa-")
```
akavel: please leave first arg to TempFile empty `""` (*["If dir is the empty string, TempFile uses the default directory for temporary files"](https://golang.org/pkg/io/ioutil/#TempFile)*)

test/e2e/util.go
```diff
@@ -189,3 +194,57 @@ func debugJSON(v interface{}) string {
...
+
+	if _, err = tmpFile.Write(out); err != nil {
+		return fmt.Errorf("Failed to write to temporary file")
+	}
+	if err := tmpFile.Close(); err != nil {
+		return fmt.Errorf("Failed to close temporary file")
+	}
+
+	// create role, role binding and service account
+	cmd := exec.Command("kubectl", "apply", "-f", tmpFile.Name())
```
akavel: instead of creating a temporary file on disk, I believe we should be able to just pass the `out` contents as in-memory stdin to kubectl:

```go
cmd := exec.Command(..., "-f", "-")
cmd.Stdin = bytes.NewReader(out)
```

scripts/e2e-test.sh
```diff
@@ -47,34 +47,32 @@ fi
...
-
-operator-sdk test local ./test/e2e --namespace "$TEST_NAMESPACE" --up-local --no-setup --verbose --go-test-flags "-tags e2e -timeout 20m"
-
-helm template deploy/helm/eunomia-operator/ \
-  --set eunomia.operator.deployment.enabled= \
-  --set eunomia.operator.namespace=$TEST_NAMESPACE | kubectl delete -f -
-
+# Eunomia setup
 helm template deploy/helm/eunomia-operator/ \
   --set eunomia.operator.image.tag=dev \
```
seanmalloy: So ... I think this means we are actually running the container image from quay.io instead of the container image built from the local branch.

seanmalloy:
@sbar95 this is now a merge conflict with `scripts/e2e-test.sh` that you will need to resolve.

scripts/e2e-test.sh
sbar95: Not in this case - in build-images.sh [freshly built images are always being tagged `dev`](https://github.com/KohlsTechnology/eunomia/blob/bab5f717b69c6599183ad4d9f7ac833859f14e74/scripts/build-images.sh#L43-L44).
I just wanted to be explicit here - when somebody reads this test script, he'll associate the `dev` tag with the development eunomia image (which he might've just modified).
Besides, the image can't be pulled from quay because of the `pullPolicy` set to `Never` one line below.

seanmalloy: Ok ... sounds good.

deploy/helm/eunomia-operator/templates/deployment.yaml
```diff
@@ -1,5 +1,5 @@
 {{- with .Values.eunomia.operator }}
-{{- if .deployment.enabled -}}
+{{- if and (.deployment.enabled) (not .deployment.nsRbacOnly) -}}
```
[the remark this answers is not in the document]
sbar95: When initially I was using bare `go test` instead of `operator-sdk test local` I tried to use the `-test.parallel=4` flag in one run to see what happens but there was no difference in test time - we need `t.Parallel()`.

To enable parallelism in `operator-sdk test local` we can use the `--go-test-flags` flag to pass the option. The command by itself sets the parallel flag in [only one case](https://github.com/operator-framework/operator-sdk/blob/v0.8.1/cmd/operator-sdk/test/local.go#L205-L207) (not ours, fortunately).

In the end `operator-sdk test local` is just [a wrapper for `go test`](https://github.com/operator-framework/operator-sdk/blob/v0.8.1/cmd/operator-sdk/test/local.go#L211-L222) - [it just uses exec](https://github.com/operator-framework/operator-sdk/blob/v0.8.1/internal/util/projutil/exec.go#L76-L86).

PR #274, opened 2020-01-30
chore: consistent shell script formatting

seanmalloy:
**Description**

Enable the shfmt utility in Travis CI to ensure that all shell scripts are consistently formatted and easy to read.

This pull request is not related to an open GitHub issue.

**Type of change**

**Checklist**

- [ ] Documentation updated

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -84,8 +84,8 @@ vet:
 # TODO: improve the command to also check scripts outside the scripts and template-processors dirs
 .PHONY: check-shfmt
 check-shfmt:
-	test -z "$(shell shfmt -i 4 -l ./scripts)"
-	test -z "$(shell shfmt -i 4 -l ./template-processors)"
+	shfmt -i 4 -d ./scripts
+	shfmt -i 4 -d ./template-processors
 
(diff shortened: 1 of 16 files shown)


Makefile
```diff
@@ -81,6 +81,12 @@ lint:
 vet:
 	VET_INPUT="$(shell go list ./... | grep -v /vendor/)"; GO111MODULE=on go vet $$VET_INPUT
 
+# TODO: improve the command to also check scripts outside the scripts and template-processors dirs
+.PHONY: check-shfmt
+check-shfmt:
+	test -z "$(shell shfmt -i 4 -l ./scripts)"
+	test -z "$(shell shfmt -i 4 -l ./template-processors)"
```
akavel: Would it still work if we [used -d flag][flag] instead of wrapping in `test -z`?

      -d        error with a diff when the formatting differs

[flag]: https://github.com/mvdan/sh/blob/30598145b5d9322bf63da248563e34bda5e3df03/cmd/shfmt/main.go#L78

seanmalloy: Yep ... good idea. Fixed.

Issue #256, opened 2020-01-17
Run Travis CI Tests Against Multiple k8s Versions


seanmalloy:
**Is your feature request related to a problem? Please describe.**

The problem is that the automated e2e tests for eunomia that are run through Travis CI are only executed against a single k8s version. The user of eunomia are more than likely running on a variety of k8s versions. Also, right now it is difficult know if updating eunomia dependencies will break eunomia(Go version, operator-sdk version, updating/adding Go modules).

**Describe the solution you'd like**

1. Update minikube to the latest version
2. Create a Travis CI build matrix to run e2e tests in parallel against multiple k8s versions
3. Use k8s versions 1.17, 1.16, 1.15, and 1.11 in the build matrix

**Describe alternatives you've considered**

Roll the dice each release and hope that it works.

**What version of eunomia are you using?**

eunomia version: v0.1.1

**Additional context**

In the future we will want to make sure that we add a build matrix for OpenShift versions too. But that is not in scope for this enhancement request.

seanmalloy:
Link to Travis CI build matrix documentation ... https://docs.travis-ci.com/user/build-matrix/

seanmalloy:
The initial test matrix has been created and minikube has been updated to the latest version. I still need to ...

- finalize the k8s versions to be used in the test matrix
- fix issue with the deploy step running for each item in the matrix

seanmalloy:
I need to do something similar to this to only run the deploy stage once.

https://docs.travis-ci.com/user/build-stages/matrix-expansion/

seanmalloy:
It works ... https://travis-ci.com/KohlsTechnology/eunomia/builds/148846276.

PR #283, opened 2020-02-04
chore: add initial TravisCI test matrix

seanmalloy:
**Description**

Run the test automation against multiple k8s versions, so that we know
which k8s versions the operator works with. A follow up change will be
made in the near future to update the minikube version which will allow
testing against new k8s versions.

This is the first part of #256.

**Type of change**

**Checklist**

- [ ] Documentation updated


seanmalloy:
Tests are running against multiple k8s versions now ...

Screen Shot 2020-02-03 at 11 04 22 PM

seanmalloy:
We can add 200 items to the build matrix, but **only 3 run in parallel**. So having more than three items in the build matrix will cause the tests to wait in a queue. We will need to select the k8s versions that we want to test carefully, or we risk having the tests run for a **very long time**.

PR #287, opened 2020-02-05
Add readiness probe

### LINKED ISSUE #210: Add kubernetes readiness check uri

In a Kubernetes deployment, it is recommended to include a readiness check for your application to let Kubernetes know when your pods are ready to accept traffic. A readiness endpoint should be created for the Eunomia operator, and it should be included in the Eunomia deployment.

A `/readiness` endpoint is created that returns `200` when the operator is able to access the Kubernetes API. This should be included in the deployment of Eunomia.

It may be possible to add this into the operator-sdk itself. It looks like there is an issue that may include a Liveness and Readiness probe.
I'm not exactly sure how it would work if it were included in the operator-sdk in the case that an end user wants to have a different check/include additional checks. I would assume it would need to be extendable/configurable for the different operator use cases.

eunomia version: v0.0.5

**Additional context**

- Issue #248  has been created to add the health/Liveness endpoint to Eunomia
- Here is the issue in the operator-sdk repo - https://github.com/operator-framework/operator-sdk/issues/1234

mkyc:
**Description**

This PR adds readinessProbe to allow kubernetes decide if requests should be routed to this pod or not. There is also separate e2e test added to verify visibility of /readyz (and already exisiting /healthz) endpoint.

Fixes #210 

**Changes**

 - adds /readyz endpoint handler in main package
 - adds readinessProbe section in helm deployment template
 - renames TEST_NAMESPACE environment variable to OPERATOR_NAMESPACE in scripts/e2e-test.sh as it was interfering with TEST_NAMESPACE variable used by operator-sdk testing framework. 
 - adds MINIKUBE_IP and WEBHOOK_PORT environment variables in scripts/e2e-test.sh to be available for new TestReadinessAndLivelinessProbes test. 
 - adds TestReadinessAndLivelinessProbes test which checks visibility of /readyz and /healthz endpoints

**Type of change**

**Checklist**

- [ok] Unit tests and e2e tests updated
- [no] Documentation updated - No related documentation. Also readinessProbe is standard mechanism.


test/e2e/probes_test.go
```diff
@@ -0,0 +1,151 @@
...
+
+func TestReadinessAndLivelinessProbes(t *testing.T) {
+	ctx := framework.NewTestCtx(t)
+	defer ctx.Cleanup()
+
+	operatorName, found := os.LookupEnv("OPERATOR_NAME")
+	if !found {
+		operatorName = "eunomia-operator"
+	}
+	operatorNamespace, found := os.LookupEnv("OPERATOR_NAMESPACE")
```
vinny-sabatini: I think the namespace should be retrieved using the method `ctx.GetNamespace()`.

Ref: https://github.com/KohlsTechnology/eunomia/commit/a41add51152126ad0e82a6e84f2b08b289890fa7

mkyc: Hmmm ... no. This `namespace, err := ctx.GetNamespace()` gets temporary namespace prepared for test, but in this test, I'm testing availability of endpoints of operator itself, so I need its namespace, not temporary one.

test/e2e/probes_test.go
```diff
@@ -0,0 +1,176 @@
...
+			endpoint:      "healthz",
+			requestBody:   "",
+			requestMethod: http.MethodGet,
+			wantCode:      http.StatusOK,
+			wantBody:      "ok",
+		},
+	}
+
+	for _, tt := range tests {
+		req, err := http.NewRequest(tt.requestMethod, fmt.Sprintf("http://%s:%d/%s", minikubeIP, nodePort, tt.endpoint), strings.NewReader(tt.requestBody))
```
akavel: `tt.requestBody` seems empty in both tests; can we remove it and just pass `nil` instead of `strings.NewReader(...)` here?

test/e2e/probes_test.go
```diff
@@ -0,0 +1,176 @@
...
+	}{
+		{
+			endpoint:      "readyz",
+			requestBody:   "",
+			requestMethod: http.MethodGet,
+			wantCode:      http.StatusOK,
+			wantBody:      "ok",
+		},
+		{
+			endpoint:      "healthz",
```
akavel: this seems to be the only field that changes between the test structs; can we remove all the other ones and just inline them in the body of the `for _, tt` loop?

test/e2e/probes_test.go
```diff
@@ -0,0 +1,176 @@
...
+			wantBody:      "ok",
+		},
+	}
+
+	for _, tt := range tests {
+		req, err := http.NewRequest(tt.requestMethod, fmt.Sprintf("http://%s:%d/%s", minikubeIP, nodePort, tt.endpoint), strings.NewReader(tt.requestBody))
+		if err != nil {
+			t.Error(err)
+		}
+		resp, err := http.DefaultClient.Do(req)
```
akavel: instead of this line and l.154, could we just use [http.Get](https://golang.org/pkg/net/http/#Get)?

test/e2e/probes_test.go
```diff
@@ -0,0 +1,176 @@
...
+		if err != nil {
+			t.Error(err)
+			continue
+		}
+		defer resp.Body.Close()
+		if tt.wantCode != resp.StatusCode {
+			t.Errorf("Returned status: %d, wanted: %d", resp.StatusCode, tt.wantCode)
```
akavel: please include the endpoint name in the error message, here and in the other ones below; personally, I typically print a subtest name using a `%q` prefix:

    t.Errorf("%q: ...", tt.endpoint, ...)

test/e2e/probes_test.go
```diff
@@ -0,0 +1,176 @@
...
+		}
+		bodyBytes, err := ioutil.ReadAll(resp.Body)
```
akavel: [optional] `body` could be enough for var name

cmd/manager/main.go
```diff
@@ -154,7 +156,19 @@ func main() {
 
 	mux.HandleFunc("/healthz", func(w http.ResponseWriter, _ *http.Request) {
 		w.WriteHeader(http.StatusOK)
-		w.Write([]byte("ok\n"))
+		w.Write([]byte("ok"))
+	})
+
+	mux.HandleFunc("/readyz", func(w http.ResponseWriter, _ *http.Request) {
+		namespaces := corev1.NamespaceList{}
+		err = mgr.GetClient().List(ctx, &client.ListOptions{}, &namespaces)
```
akavel: Why do we check namespaces in particular? Could you provide a comment with an explanation in the code of this func? I'm not (yet? 😜) trying to argue with this choice, rather trying to understand it in the first place.

deploy/helm/eunomia-operator/templates/deployment.yaml
```diff
@@ -37,6 +37,10 @@ spec:
               port: 8080
             initialDelaySeconds: 30
             periodSeconds: 5
+          readinessProbe:
+            httpGet:
+              path: /readyz
+              port: 8080
```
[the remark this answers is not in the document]
mkyc: Nope. Firstly here are defaults: https://github.com/kubernetes/kubernetes/blob/master/pkg/apis/core/v1/defaults_test.go#L1819-L1824 and secondly readinessProbe instruct Kubernetes if it should route traffic to this pod, so there is no problem if initial delay is 0 and readiness probe fails during startup of operator.

test/e2e/probes_test.go
```diff
@@ -0,0 +1,176 @@
...
+		operatorNamespace = "test-eunomia-operator"
+	}
+	minikubeIP, found := os.LookupEnv("MINIKUBE_IP")
+	if !found {
+		minikubeIP = "localhost"
+	}
+	webHookPort, found := os.LookupEnv("WEBHOOK_PORT")
+	if !found {
+		webHookPort = "8080"
+	}
```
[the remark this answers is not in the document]
mkyc: You are so right! I didn't think it through.

test/e2e/probes_test.go
mkyc: I believe it's good to see clearly what test is trying to verify. I removed `requestBody` and `requestMethod` but I think I'd keep `want...` things. If you think that this is important to remove those as well, let me know.

cmd/manager/main.go
seanmalloy: I'm wondering if instead of list namespaces we should list GitOpsConfigs. What do you think?

seanmalloy:
@mkyc I had one inline comment/question [here](https://github.com/KohlsTechnology/eunomia/pull/287/files#r376041604). Let me know what you think.

Other than that I think we just need @vinny-sabatini and @akavel to review one more time.

cmd/manager/main.go
mkyc: Generally I do not see benefit of listing GitOpsConfigs over Namespaces. Virtually it can be anything (e.g.: ConfigMaps, Pods, etc.) because we are only interested here in if there is error while calling API or not. We do not even care if there actually was something returned, but if we were, Namespace is still the most obvious choice, because there would always be some regardless of level of access Pod has. We could add here check if returned list is not empty and that would make some sense. But let's remember that this operation is frequent, and not cached via informer mechanism (which is essential for this check). Also I didn't want to produce heavy load of code there. 

Let me know what you think :)

test/e2e/probes_test.go
```diff
@@ -40,27 +40,31 @@ func TestReadinessAndLivelinessProbes(t *testing.T) {
...
 	}
 	webHookPortInt, err := strconv.Atoi(webHookPort)
 	if err != nil {
-		t.Error(err)
+		t.Fatal(err)
 	}
 
 	t.Logf("minikube IP: %s", minikubeIP)
 
+	// Aim of this test is to validate that required endpoints are there and are accessible. To achieve this,
```
akavel:
```suggestion
	// The aim of this test is to validate that the required endpoints are there and are accessible. To achieve this,
```

test/e2e/probes_test.go
```diff
@@ -40,27 +40,31 @@ func TestReadinessAndLivelinessProbes(t *testing.T) {
...
+	// we need to expose deployment externally outside of minikube as this test runs outside of it. To ensure
```
akavel:
```suggestion
	// we need to expose the deployment externally outside of minikube, as this test runs outside of it. To ensure
```

test/e2e/probes_test.go
```diff
@@ -40,27 +40,31 @@ func TestReadinessAndLivelinessProbes(t *testing.T) {
...
+	// that this is done correctly Service with Type: "NodePort" is created here, to expose operator webhook
```
akavel:
```suggestion
	// that this is done correctly, a Service with Type: "NodePort" is created here, to expose the operator's webhook
```

test/e2e/probes_test.go
```diff
@@ -105,72 +109,52 @@ func TestReadinessAndLivelinessProbes(t *testing.T) {
...
 		}
 		defer resp.Body.Close()
 		if tt.wantCode != resp.StatusCode {
-			t.Errorf("Returned status: %d, wanted: %d", resp.StatusCode, tt.wantCode)
+			t.Errorf("%q: returned status: %d, wanted: %d", tt.endpoint, resp.StatusCode, tt.wantCode)
 		}
-		bodyBytes, err := ioutil.ReadAll(resp.Body)
+		body, err := ioutil.ReadAll(resp.Body)
 		if err != nil {
 			t.Error(err)
```
akavel:
```suggestion
			t.Errorf("%q: %s", tt.endpoint, err)
```

test/e2e/probes_test.go
```diff
@@ -105,72 +109,52 @@ func TestReadinessAndLivelinessProbes(t *testing.T) {
...
 
 	for _, tt := range tests {
-		req, err := http.NewRequest(tt.requestMethod, fmt.Sprintf("http://%s:%d/%s", minikubeIP, nodePort, tt.endpoint), strings.NewReader(tt.requestBody))
-		if err != nil {
-			t.Error(err)
-		}
-		resp, err := http.DefaultClient.Do(req)
+		resp, err := http.Get(fmt.Sprintf("http://%s:%d/%s", minikubeIP, nodePort, tt.endpoint))
 		if err != nil {
 			t.Error(err)
```
akavel:
```suggestion
			t.Errorf("%q: %s", tt.endpoint, err)
```

cmd/manager/main.go
akavel: @mkyc thanks for the comment with explanation in code, now I understand and it makes sense to me, with my limited knowledge of Kubernetes.

scripts/e2e-test.sh
```diff
@@ -57,8 +57,9 @@ GOOS=linux make e2e-test-images
 # shellcheck disable=SC2155
 export MINIKUBE_IP=$(minikube ip)
 
-# Define webhook port
-export WEBHOOK_PORT=8080
+# TestReadinessAndLivelinessProbes is accessing operator via newly created service
+# and it needs to know what is the port to connect to.
+export OPERATOR_WEBHOOK_PORT=8080
```
akavel: As we talked, it would be cool if you could also add some info where's the particular value 8080 coming from.

test/e2e/probes_test.go
akavel: [optional] I would personally remove as we talked, but I understand what you mean and leave the choice to you.

test/e2e/probes_test.go
vinny-sabatini: Ahhhh yes, misunderstanding on my part. Thanks!

cmd/manager/main.go
seanmalloy: We can try using namespace and see how well it works.

PR #293, opened 2020-02-11
handler: add Ref check to WebhookHandler

### LINKED ISSUE #252: Webhook code should filter for a ref/branch name match before kicking off

eunomia version:
v.0.1.0
yes

$ kubectl version

</pre></details>

**What did you do?**
done a git push to a branch ref that is not configured in a GitOpsConfig CR

Eunomia should not kickoff any jobs when the webhook payload indicates 'git push' was done to a branch name different to a ref name specified in the GitOpsConfig CR deployed in the cluster

**What did you see instead?**
Eunomia kicks-off a job/pod upon receiving any 'git push' webhook, no matter if the GitOpsConfig CR specifed ref matches the one from webhook

rhalat:
**Description**

GitHub webhook is triggered on all Push events and without checking 
branch in WebhookHandler new job is created every time there is push to 
repo, no matter what branch. In this PR I added check, so job is created 
only when URI and Ref matches in push event and GitOpsConfig CR.

Additionally it's fixing Travis CI that was failing on hello-world-yaml-cr1 test by increasing timeout.

Fixes #252

**Type of change**

**Checklist**

- [x] Documentation updated

### DIFF
--- a/pkg/handler/handler.go
+++ b/pkg/handler/handler.go
@@ -76,7 +76,7 @@ func WebhookHandler(w http.ResponseWriter, r *http.Request, reconciler gitopscon
 					continue
 				}
 				// if the repo URL do not correspond continue
-				if !repoURLMatch(&instance, e) {
+				if !repoURLAndRefMatch(&instance, e) {
 					continue
 				}
 				targetList.Items = append(targetList.Items, instance)
@@ -117,8 +117,12 @@ func WebhookHandler(w http.ResponseWriter, r *http.Request, reconciler gitopscon
 	log.Info("webhook handling concluded correctly")
 }
 
-func repoURLMatch(instance *gitopsv1alpha1.GitOpsConfig, event *github.PushEvent) bool {
-	return strings.Contains(instance.Spec.TemplateSource.URI, *event.Repo.FullName) || strings.Contains(instance.Spec.ParameterSource.URI, *event.Repo.FullName)
+func repoURLAndRefMatch(instance *gitopsv1alpha1.GitOpsConfig, event *github.PushEvent) bool {
+	return event.Repo != nil && event.Repo.FullName != nil && event.Ref != nil &&
+		((strings.Contains(instance.Spec.TemplateSource.URI, *event.Repo.FullName) &&
+			instance.Spec.TemplateSource.Ref == strings.TrimPrefix(*event.Ref, "refs/heads/")) ||
+			(strings.Contains(instance.Spec.ParameterSource.URI, *event.Repo.FullName) &&
+				instance.Spec.ParameterSource.Ref == strings.TrimPrefix(*event.Ref, "refs/heads/")))
 }
 
 func getWebhookSecret(instance *gitopsv1alpha1.GitOpsConfig) string {
--- a/pkg/handler/handler_test.go
+++ b/pkg/handler/handler_test.go
@@ -1,0 +1,150 @@
+/*
+Copyright 2020 Kohl's Department Stores, Inc.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+	http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package handler
+
+import (
+	"testing"
+
+	"github.com/google/go-github/github"
+
+	gitopsv1alpha1 "github.com/KohlsTechnology/eunomia/pkg/apis/eunomia/v1alpha1"
+)
+
+func newstring(v string) *string { return &v }
+
+func TestRepoURLAndRefMatch(t *testing.T) {
+
+	defaultGit := gitopsv1alpha1.GitConfig{
+		URI: "https://github.com/kohlstechnology/eunomia",
+		Ref: "master",
+	}
+
+	tests := []struct {
+		comment string
+		spec    gitopsv1alpha1.GitOpsConfigSpec
+		event   github.PushEvent
+		want    bool
+	}{
+		{
+			comment: "full match - branch",
+			spec:    gitopsv1alpha1.GitOpsConfigSpec{TemplateSource: defaultGit, ParameterSource: defaultGit},
+			event: github.PushEvent{
+				Ref: newstring("master"),
+				Repo: &github.PushEventRepository{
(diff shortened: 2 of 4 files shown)


pkg/handler/handler.go
```diff
@@ -117,8 +117,11 @@ func WebhookHandler(w http.ResponseWriter, r *http.Request, reconciler gitopscon
 	log.Info("webhook handling concluded correctly")
 }
 
-func repoURLMatch(instance *gitopsv1alpha1.GitOpsConfig, event *github.PushEvent) bool {
-	return strings.Contains(instance.Spec.TemplateSource.URI, *event.Repo.FullName) || strings.Contains(instance.Spec.ParameterSource.URI, *event.Repo.FullName)
+func repoURLAndRefMatch(instance *gitopsv1alpha1.GitOpsConfig, event *github.PushEvent) bool {
```
akavel: Please add a unit test for this func, with at least 1 positive and 1 negative test case. Some helpful links for learning how to write unit tests in Go in a standard way:

- https://golang.org/doc/code.html#Testing
- https://dave.cheney.net/2019/05/07/prefer-table-driven-tests

Suggested initial template for the test function:

```
func TestRepoURLAndRefMatch(t *testing.T) {
    tests := []struct{
        comment  string
        instance *gitopsv1alpha1.GitOpsConfig
        event    *github.PushEvent
        want     bool
    }{
        {
             comment: "full match",
             ...
             want: true,
        },
        {
             comment: "parameterSource.Ref differs",
             ...
             want: false,
        },
        {
             comment: "parameterSource.URI differs",
             ...
             want: false,
        },
    }

    for _, tt := range tests {
         result := repoURLAndRefMatch(tt.instance, tt.event)
         if result != tt.want {
             t.Errorf("%q: expected %v, got %v", tt.comment, tt.want, match)
         }
    }
}
```

seanmalloy:
@rhalat what if `.Spec.ParameterSource.Ref` or `.Spec.TemplateSource.Ref` is not a branch? How does this code handle a git tag or git sha1?

rhalat:
@seanmalloy is it supposed to handle something different than branches? When it comes to sha1, I believe it won't trigger job and when it comes to tags we can think of it and I can include it in unit tests, but what in case where branch and tag have same names? I GitOpsConfig CR we are not distinguish them.

akavel:
@seanmalloy @rhalat Maybe we could just tell users to set Ref to `refs/tags/foobar` when they want to trigger on `foobar` tag? I suppose this probably already works now with this PR, and would also help us avoid name conflicts?

seanmalloy:
@rhalat please make the code work with git branches and git tags. We can worry about the git SHA1 use case at some point in the future.

As @akavel mentioned maybe we just need to update the docs for now, so people know how to use the webhook trigger with both branches and tags.

rhalat:
I updated README.md and I'm close to finish with unit tests.

pkg/handler/handler_test.go
```diff
@@ -0,0 +1,151 @@
...
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package handler
+
+import (
+	gitopsv1alpha1 "github.com/KohlsTechnology/eunomia/pkg/apis/eunomia/v1alpha1"
+	"github.com/google/go-github/github"
+	"testing"
```
akavel: Please split imports into groups, separated by empty line:

1. standard library imports
2. third-party imports
3. project-internal imports

pkg/handler/handler_test.go
```diff
@@ -0,0 +1,151 @@
...
+			want: true,
+		},
+		{
+			comment: "Ref differs",
+			spec: gitopsv1alpha1.GitOpsConfigSpec{
+				TemplateSource: defaultGit,
+				ParameterSource: gitopsv1alpha1.GitConfig{
+					URI: defaultGit.URI,
+					Ref: defaultGit.Ref,
+				},
```
akavel: I think you can use `ParameterSource: defaultGit` here?

rhalat: Yes, thank you!

pkg/handler/handler_test.go
rhalat: Done

pkg/handler/handler.go
```diff
@@ -117,8 +117,11 @@ func WebhookHandler(w http.ResponseWriter, r *http.Request, reconciler gitopscon
...
+	return (strings.Contains(instance.Spec.TemplateSource.URI, *event.Repo.FullName) &&
+				 instance.Spec.TemplateSource.Ref == strings.TrimPrefix(*event.Ref, "refs/heads/")) ||
```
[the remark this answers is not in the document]
rhalat: There is no much information about that. All what I could find is webhook payload example:
https://developer.github.com/v3/activity/events/types/#webhook-payload-example-31

and topic from 2 weeks ago in GitHub community that it's not well documented:
https://github.community/t5/GitHub-API-Development-and/Push-webhook-payload-is-inexplicable/td-p/45416

But honestly I don't think that either `Ref` or `FullName` can be `nil`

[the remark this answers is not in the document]
rhalat: I added code to check if `event.Repo`, `event.Repo.FullName` and `event.Ref` are not `nil` in latest commit

pkg/handler/handler.go
```diff
@@ -118,8 +118,9 @@ func WebhookHandler(w http.ResponseWriter, r *http.Request, reconciler gitopscon
 }
 
 func repoURLAndRefMatch(instance *gitopsv1alpha1.GitOpsConfig, event *github.PushEvent) bool {
-	return (strings.Contains(instance.Spec.TemplateSource.URI, *event.Repo.FullName) &&
-		instance.Spec.TemplateSource.Ref == strings.TrimPrefix(*event.Ref, "refs/heads/")) ||
+	return event.Repo != nil && event.Repo.FullName != nil && event.Ref != nil &&
+		(strings.Contains(instance.Spec.TemplateSource.URI, *event.Repo.FullName) &&
```
akavel: I think the `||` block needs to be in an extra pair of parentheses...

NIL_CHECKS && ((CONTAINS_CHECK && REF_CHECK) || (CONTAINS_CHECK && REF_CHECK))

otherwise I think the precedence rules will group all `&&`s together

rhalat: Fixed

akavel:
@rhalat Please add an extra commit increasing the timeout in e2e-test.sh for hello-world-yaml-cr1. Please make sure the commit  message explains the reasons: from CI logs, we see that one run barely completes in time:

https://travis-ci.com/KohlsTechnology/eunomia/jobs/288138640#L2471

so the failure in the other run will probably disappear if we increase timeout:

https://travis-ci.com/KohlsTechnology/eunomia/jobs/288138641#L2469

As a side note, personally, I hate that the tests are taking so long now; I hope we'll find some way(s) to speed them up at some point.

PR #295, opened 2020-02-13
Only Run Travis CI Deploy Step Once

### LINKED ISSUE #256: Run Travis CI Tests Against Multiple k8s Versions

seanmalloy:
**Description**
When the original build matrix was added to enable testing multiple k8s
versions in Travis CI the deploy would also run for each k8s version in
the build matrix. This changes makes the deploy step only run once.

See below Travis CI docs for details.

https://docs.travis-ci.com/user/build-stages/#build-stages-and-deployments

Fixes #256 

**Type of change**

**Checklist**

- [ ] Documentation updated

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -67,9 +67,9 @@ jobs:
 stages:
   - test
   - name: deploy
-    if: type IN (push) AND branch = master
+    if: type = push AND branch = master AND fork = false
   - name: deploy-release
-    if: tag =~ /^v\d+\.\d+\.\d+$/
+    if: tag =~ /^v\d+\.\d+\.\d+$/ AND fork = false
 
 after_success:
   - bash <(curl -s https://codecov.io/bash)
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,8 +15,8 @@ env:
   # NOTE: We can add 200 items to the build matrix, but only 3 run in parallel. So having more than three
   # items in the build matrix will cause the tests to wait in a queue. We will need to select the k8s versions
   # that we want to test carefully, or we risk having the tests run for a very long time.
-  - OPERATOR_SDK_VERSION=v0.8.1 KUBE_VERSION=v1.14.2 MINIKUBE_VERSION=v1.6.2 SHFMT_VERSION=v3.0.1
-  - OPERATOR_SDK_VERSION=v0.8.1 KUBE_VERSION=v1.13.2 MINIKUBE_VERSION=v1.6.2 SHFMT_VERSION=v3.0.1
+  - OPERATOR_SDK_VERSION=v0.8.1 KUBE_VERSION=v1.15.10 MINIKUBE_VERSION=v1.6.2 SHFMT_VERSION=v3.0.1
+  - OPERATOR_SDK_VERSION=v0.8.1 KUBE_VERSION=v1.16.7 MINIKUBE_VERSION=v1.6.2 SHFMT_VERSION=v3.0.1
 
 before_install:
   - travis_retry go mod vendor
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -53,19 +53,24 @@ before_script:
 script:
   - make test-dirty test
 
+jobs:
+  include:
+    - stage: deploy
+      env:
+        - OPERATOR_SDK_VERSION=v0.8.1 KUBE_VERSION=v1.15.10 MINIKUBE_VERSION=v1.6.2 SHFMT_VERSION=v3.0.1
```
asinkavets: @seanmalloy why we're not use vars from the beginning of this yml? Smth like `OPERATOR_SDK_VERSION=${OPERATOR_SDK_VERSION}`

.travis.yml
```diff
@@ -53,19 +53,24 @@ before_script:
...
+      script: make travis-release
+
+stages:
+  - test
+  - name: deploy
+    if: type IN (push) AND branch = master
+  - name: deploy-release
+    if: tag =~ /^v\d+\.\d+\.\d+$/
```
akavel: 1. [optional] The Travis docs are somewhat confusing to me, but looking at the ["Conditional Jobs"](https://docs.travis-ci.com/user/conditional-builds-stages-jobs/#conditional-jobs) section, I'm kinda thinking maybe we can put the `if` blocks already in the `jobs: include:` block above? and then not have to add the separate `stages:` block? that said, I think I'd try testing this stuff on some separate sandbox repo first anyway...
  2. Shouldn't we also carry over an `AND repo = KohlsTechnology/eunomia` condition, for both `deploy` and `deploy-release`? I would guess this was probably intended to avoid `make travis-deploy-images` and `make travis-release` being called during CI in forked repos? Actually, there [seems to exist][fork] a `fork = false` condition, maybe it'll make the intention clearer:

```suggestion
    if: tag =~ /^v\d+\.\d+\.\d+$/ AND fork = false
```

[fork]: https://docs.travis-ci.com/user/conditions-v1#examples

.travis.yml
```diff
@@ -53,19 +53,24 @@ before_script:
...
+      script: make travis-deploy-images
+    - stage: deploy-release
+      env:
+        - OPERATOR_SDK_VERSION=v0.8.1 KUBE_VERSION=v1.15.10 MINIKUBE_VERSION=v1.6.2 SHFMT_VERSION=v3.0.1
+      script: make travis-release
+
+stages:
+  - test
+  - name: deploy
+    if: type IN (push) AND branch = master
```
akavel:
```suggestion
    if: type = push AND branch = master AND fork = false
```

.travis.yml
```diff
@@ -15,8 +15,8 @@ env:
   # NOTE: We can add 200 items to the build matrix, but only 3 run in parallel. So having more than three
   # items in the build matrix will cause the tests to wait in a queue. We will need to select the k8s versions
   # that we want to test carefully, or we risk having the tests run for a very long time.
-  - OPERATOR_SDK_VERSION=v0.8.1 KUBE_VERSION=v1.14.2 MINIKUBE_VERSION=v1.6.2 SHFMT_VERSION=v3.0.1
-  - OPERATOR_SDK_VERSION=v0.8.1 KUBE_VERSION=v1.13.2 MINIKUBE_VERSION=v1.6.2 SHFMT_VERSION=v3.0.1
+  - OPERATOR_SDK_VERSION=v0.8.1 KUBE_VERSION=v1.15.10 MINIKUBE_VERSION=v1.6.2 SHFMT_VERSION=v3.0.1
+  - OPERATOR_SDK_VERSION=v0.8.1 KUBE_VERSION=v1.16.7 MINIKUBE_VERSION=v1.6.2 SHFMT_VERSION=v3.0.1
```
akavel: Please also add an explanation in the commit message why we're upgrading the kube versions as well

.travis.yml
seanmalloy: Without setting the environment variables the `deploy` runs multiple times. Once for each matrix dimension(item?). So we end end running the `deploy` twice, once for each `KUBE_VERSION` in the matrix.

It might be possible to use `jobs.include` in make this cleaner. But I think we need to do something explicit in the YAML to only run the `deploy` a single time.

.travis.yml
seanmalloy: I mocked up a `.travis.yml` over here for testing ... https://github.com/seanmalloy/travis-testing. This is how I made sure it all works.

I'll add the additional fork condition and re-test it.

seanmalloy: Fixed.

.travis.yml
seanmalloy: Fixed commit message.

.travis.yml
seanmalloy: Fixed.

.travis.yml
seanmalloy: I'd prefer to keep this as is because I know it is working.

seanmalloy: Let me know what you think.

seanmalloy: I re-read the doc  ["Build Stages and Build Matrix Expansion"](https://docs.travis-ci.com/user/build-stages/#build-stages-and-build-matrix-expansion). The implicit build matrix include feature does not work in our case. This only works when specifying the language version, which is in our case Go(see line 9).

We have to explicitly include `env` for the deploy jobs because we also have environment variables in our matrix too(see lines 14 to 19).

Summary: I believe we have to do it the way I have proposed in this PR, or the deploy jobs will run once for each `KUBE_VERSION` in the matrix.

PR #301, opened 2020-02-18
Allow user to pass empty TemplateSource ContextDir

### LINKED ISSUE #276: Base Template Processor Fails To Apply 'contextDir' With Zero YAML Files

eunomia version: v0.1.1

Yes.

$ kubectl version
OCP v3.11
</pre></details>

**What did you do?**
I applied a simple GitOpsConifg CR to my OpenShift 3.11 cluster. The `contextDir` did not contain any YAML files.

Example GitOpConfig CR ...
```
---
apiVersion: eunomia.kohls.io/v1alpha1
kind: GitOpsConfig
metadata:
  name: namespace-poc
spec:
  templateSource:
    uri: https://github.example.com/operations/openshift-namespaces
    ref: bug-tester
    contextDir: mycluster/namespaces
    secretRef: eunomia-runner-github
  triggers:
  - type: Webhook
  serviceAccountRef: eunomia-runner
  templateProcessorImage: docker-registry.default.svc:5000/default/eunomia-base:latest
  resourceHandlingMode: Apply
  resourceDeletionMode: Delete
```

I expect the eunomia base template processor image to be able to successfully apply a `contextDir` that contains zero YAML files.

**What did you see instead?**
The base template process fails with the below error ...
```
+ echo export SERVICE_CA_BUNDLE=
+ getNamespace
++ cat /var/run/secrets/kubernetes.io/serviceaccount/namespace
+ echo export NAMESPACE=xpaas-eunomia
+ source /tmp/envs.sh
++ export CA_BUNDLE=
++ CA_BUNDLE=
++ export SERVICE_CA_BUNDLE=
++ SERVICE_CA_BUNDLE=
++ export NAMESPACE=xpaas-eunomia
++ NAMESPACE=xpaas-eunomia
+ /usr/local/bin/processParameters.sh
+ echo 'Processing Parameters'
+ FOLDERS=
+ '[' -e /git/parameters/./hierarchy.lst ']'
+ FOLDERS=/git/parameters/.
+ '[' -n /git/parameters/. ']'
+ VALUES_FILE=/tmp/eunomia_values_processed1.yaml
+ echo ---
Processing Parameters
+ for DIR in ${FOLDERS}
+ echo 'Processing files in /git/parameters/.'
Processing files in /git/parameters/.
++ find /git/parameters/. -maxdepth 1 -name '*.json' -o -name '*.yaml' -o -name '*.yml'
+ YAML_FILES=/git/parameters/./cr.yaml
+ '[' '!' -z /git/parameters/./cr.yaml ']'
+ goyq merge -i -x /tmp/eunomia_values_processed1.yaml /git/parameters/./cr.yaml
+ '[' -e /tmp/eunomia_values_processed1.yaml ']'
+ envsubst
+ /usr/local/bin/processTemplates.sh
Processing Templates
+ echo Processing Templates
+ cp -R '/git/templates/mycluster/namespaces/*' /git/manifests/
cp: cannot stat '/git/templates/mycluster/namespaces/*': No such file or directory
```

sbar95:
**Description**

Check if CLONED_TEMPLATE_GIT_DIR exists while processing template files.
This resolves an issue occuring when user specifies an empty parameters
directory in GitOpsConfig in a remote repository. Such directory won't
exists because of the way git works - it doesn't track empty
directories. If user creted an empty directory with .gitkeep (or
similiar) file explicitly, serve such case too by printing a message in
pod logs (do not fail in such a case, user knows what he is doing).

Reference:
https://git.wiki.kernel.org/index.php/Git_FAQ#Can_I_add_empty_directories.3F

**ALSO:**
Double-check if the existing job is blocked because of bad image. Add
checking if there is only one active pod of a job, which occurs when
nonexistent or wrong image blocks the job. Thanks to it only such case
will fall under the if code block.
Before, if there was a "normal" fail of a job (i.e. code inside
container exited 1), and there were a few failed pods, this would fall
under the if block, and jobContainerStatus function would return an
error because the job had a few pods.

Fixes #276 

**Type of change**

**Checklist**

- [ ] Documentation updated (not required for this PR)

### DIFF
--- a/template-processors/base/bin/processTemplates.sh
+++ b/template-processors/base/bin/processTemplates.sh
@@ -18,5 +18,4 @@ set -euxo pipefail
 
 echo Processing Templates
 
-# shellcheck disable=SC2086
-cp -R "$CLONED_TEMPLATE_GIT_DIR/"* "$MANIFEST_DIR"/
+cp -R "${CLONED_TEMPLATE_GIT_DIR}/." "${MANIFEST_DIR}/"
--- a/template-processors/base/bin/resourceManager.sh
+++ b/template-processors/base/bin/resourceManager.sh
@@ -44,7 +44,7 @@ function addLabels() {
     local timestamp="$2"
     local tmpdir="$(mktemp -d)"
     # shellcheck disable=SC2044
-    for file in $(find "$MANIFEST_DIR" -iregex '.*\.\(ya?ml\|json\)'); do
+    for file in $(find "$MANIFEST_DIR" -regextype posix-extended -iregex '.*\.(ya?ml|json)'); do
         cat "$file" |
             yq -y -s "map(select(.!=null)|setpath([\"metadata\",\"labels\",\"$TAG_OWNER\"]; \"$owner\"))|.[]" |
             yq -y -s "map(select(.!=null)|setpath([\"metadata\",\"labels\",\"$TAG_APPLIED\"]; \"$timestamp\"))|.[]" \
@@ -80,6 +80,16 @@ function deleteByOldLabels() {
 function createUpdateResources() {
     local owner="$1"
     local timestamp="$(date +%s)"
+    # Check if directory contains only hidden files like .gitkeep, or .gitignore.
+    # This would mean that user purposefully wanted to track an empty directory in git.
+    # https://git.wiki.kernel.org/index.php/Git_FAQ#Can_I_add_empty_directories.3F
+    if [[ -z $(ls "${MANIFEST_DIR}") ]]; then
+        echo "Manifest directory empty, skipping"
+        return
+    elif [[ -z $(find "$MANIFEST_DIR" -regextype posix-extended -iregex '.*\.(ya?ml|json)') ]]; then
+        echo "ERROR - no files with .yaml, .yml, or .json extension in manifest directory"
+        exit 1
+    fi
     case "$CREATE_MODE" in
     Apply)
         addLabels "$owner" "$timestamp"
--- a/test/e2e/issue276_test.go
+++ b/test/e2e/issue276_test.go
@@ -1,0 +1,263 @@
+// +build e2e
+
+/*
+Copyright 2020 Kohl's Department Stores, Inc.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+	http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package e2e
+
+import (
+	"context"
+	"fmt"
+	"os"
+	"strings"
+	"testing"
+	"time"
+
+	framework "github.com/operator-framework/operator-sdk/pkg/test"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/wait"
+
+	"github.com/KohlsTechnology/eunomia/pkg/apis"
+	gitopsv1alpha1 "github.com/KohlsTechnology/eunomia/pkg/apis/eunomia/v1alpha1"
+)
+
+// TestIssue276NoTemplatesDir verifies that job's pod fails in such a
+// way that it prints a custom error message to output when user passes
+// nonexistent TemplateSource ContextDir. Then, the test also verifies
+// that the Custom Recource is successfully deleted
+func TestIssue276NoTemplatesDir(t *testing.T) {
+	if testing.Short() {
+		// FIXME: as of writing this test, "backoffLimit" in job.yaml is set to 4,
+		// which means eunomia will wait to launch deletion Job until 5 Pod retries
+		// fail, eventually triggering the origninal Job's failure; the back-off
+		// time between the runs is unfortunately exponential and non-configurable,
+		// which makes this test awfully long. Try to at least make it possible to
+		// run in parallel with other tests.
+		t.Skip("This test currently takes minutes to run, because of exponential backoff in kubernetes")
+	}
+
+	ctx := framework.NewTestCtx(t)
+	defer ctx.Cleanup()
+
+	namespace, err := ctx.GetNamespace()
+	if err != nil {
+		t.Fatalf("could not get namespace: %v", err)
+	}
+	if err = SetupRbacInNamespace(namespace); err != nil {
+		t.Error(err)
+	}
+
+	defer DumpJobsLogsOnError(t, framework.Global, namespace)
+	err = framework.AddToFrameworkScheme(apis.AddToScheme, &gitopsv1alpha1.GitOpsConfigList{})
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	eunomiaURI, found := os.LookupEnv("EUNOMIA_URI")
+	if !found {
+		eunomiaURI = "https://github.com/kohlstechnology/eunomia"
+	}
+	eunomiaRef, found := os.LookupEnv("EUNOMIA_REF")
+	if !found {
+		eunomiaRef = "master"
+	}
+
+	// Step 1: create a CR with a nonexistent TemplateSource ContextDir
+
+	const noDir = "test/e2e/testdata/no-directory"
+	gitops := &gitopsv1alpha1.GitOpsConfig{
+		TypeMeta: metav1.TypeMeta{
+			Kind:       "GitOpsConfig",
+			APIVersion: "eunomia.kohls.io/v1alpha1",
+		},
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "gitops-issue276",
+			Namespace: namespace,
+			Finalizers: []string{
+				"gitopsconfig.eunomia.kohls.io/finalizer",
+			},
+		},
+		Spec: gitopsv1alpha1.GitOpsConfigSpec{
+			TemplateSource: gitopsv1alpha1.GitConfig{
+				URI:        eunomiaURI,
+				Ref:        eunomiaRef,
+				ContextDir: noDir,
+			},
+			ParameterSource: gitopsv1alpha1.GitConfig{
+				URI:        eunomiaURI,
+				Ref:        eunomiaRef,
+				ContextDir: "test/e2e/testdata/empty-yaml",
+			},
+			Triggers: []gitopsv1alpha1.GitOpsTrigger{
+				{Type: "Change"},
+			},
+			TemplateProcessorImage: "quay.io/kohlstechnology/eunomia-base:dev",
+			ResourceHandlingMode:   "Apply",
+			ResourceDeletionMode:   "Delete",
+			ServiceAccountRef:      "eunomia-operator",
+		},
+	}
+	gitops.Annotations = map[string]string{"gitopsconfig.eunomia.kohls.io/initialized": "true"}
+
+	err = framework.Global.Client.Create(context.TODO(), gitops, &framework.CleanupOptions{TestContext: ctx, Timeout: timeout, RetryInterval: retryInterval})
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	// Step 2: Wait until Job's pod fails and check if it printed a clear error message
+
+	const name = "gitopsconfig-gitops-issue276-"
+	err = wait.Poll(retryInterval, timeout, func() (done bool, err error) {
+		pod, err := GetPod(namespace, name, "quay.io/kohlstechnology/eunomia-base:dev", framework.Global.KubeClient)
+		switch {
+		case apierrors.IsNotFound(err):
+			t.Logf("Waiting for availability of %s pod", name)
+			return false, nil
+		case err != nil:
+			return false, err
+		case pod != nil && pod.Status.Phase == "Failed":
+			logs, err := GetPodLogs(pod, framework.Global.KubeClient)
+			if err != nil {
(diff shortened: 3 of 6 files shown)


test/e2e/issue276_test.go
```diff
@@ -0,0 +1,242 @@
...
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/wait"
+
+	"github.com/KohlsTechnology/eunomia/pkg/apis"
+	gitopsv1alpha1 "github.com/KohlsTechnology/eunomia/pkg/apis/eunomia/v1alpha1"
+)
+
+// TestIssue276NoTemplatesDir verifies that job's pod fails in a proper
+// way when user passes nonexistent TemplateSource ContextDir
```
akavel: Please explain "fails in a proper way"

test/e2e/issue276_test.go
```diff
@@ -0,0 +1,242 @@
...
+			TemplateProcessorImage: "quay.io/kohlstechnology/eunomia-base:dev",
+			ResourceHandlingMode:   "Apply",
+			ResourceDeletionMode:   "Delete",
+			ServiceAccountRef:      "eunomia-operator",
+		},
+	}
+	gitops.Annotations = map[string]string{"gitopsconfig.eunomia.kohls.io/initialized": "true"}
+
+	// Triple timeout.
+	// FIXME: as of writing this test, "backoffLimit" in job.yaml is set to 4,
```
akavel: I don't understand: what does backoffLimit have to do with CleanupOptions?

test/e2e/issue276_test.go
```diff
@@ -0,0 +1,242 @@
...
+			return false, nil
+		case err != nil:
+			return false, err
+		case pod != nil && pod.Status.Phase == "Failed":
+			logs, err := GetPodLogs(pod, framework.Global.KubeClient)
+			if err != nil {
+				t.Fatal(err)
+			}
+			if !strings.Contains(logs, fmt.Sprintf("ERROR - directory %s does not exist in the remote repository", noDir)) {
+				t.Fatal(err)
```
akavel: err will be nil here, no? should we print some more descriptive info here? did you verify if the test fails when you revert the code to the version before fix?

test/e2e/issue276_test.go
```diff
@@ -0,0 +1,242 @@
...
+			return false, nil
+		}
+	})
+	if err != nil {
+		t.Error(err)
+	}
+}
+
+// TestIssue276EmptyTemplatesDir verifies that job succeeds when user
+// passes empty (in git repository sense) TemplateSource ContextDir
```
akavel: I don't like the phrase "empty (in git repository sense)" phrase, I find it confusing.  I would suggest rephrasing it, for example:
```suggestion
// passes a TemplateSource directory containing no resource files, but only a single ".gitkeep" file
```

test/e2e/issue276_test.go
```diff
@@ -0,0 +1,242 @@
...
+	eunomiaURI, found := os.LookupEnv("EUNOMIA_URI")
+	if !found {
+		eunomiaURI = "https://github.com/kohlstechnology/eunomia"
+	}
+	eunomiaRef, found := os.LookupEnv("EUNOMIA_REF")
+	if !found {
+		eunomiaRef = "master"
+	}
+
+	// Step 1: create a CR with an empty (in git repository sense) TemplateSource ContextDir
```
akavel: similar as above

test/e2e/issue276_test.go
```diff
@@ -0,0 +1,242 @@
...
+		switch {
+		case apierrors.IsNotFound(err):
+			t.Logf("Waiting for availability of %s pod", name)
+			return false, nil
+		case err != nil:
+			return false, err
+		case pod != nil && pod.Status.Phase == "Succeeded":
+			return true, nil
+		case pod != nil:
+			t.Logf("Waiting for error in pod %s; status: %s", pod.Name, debugJSON(pod.Status))
```
akavel: I don't think we're waiting for error here?

test/e2e/issue276_test.go
```diff
@@ -0,0 +1,242 @@
...
+	gitops.Annotations = map[string]string{"gitopsconfig.eunomia.kohls.io/initialized": "true"}
+
+	err = framework.Global.Client.Create(context.TODO(), gitops, &framework.CleanupOptions{TestContext: ctx, Timeout: timeout, RetryInterval: retryInterval})
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	// Step 2: Wait until Job succeeds
+
+	err = wait.Poll(retryInterval, timeout, func() (done bool, err error) {
```
akavel: Can we just use WaitForPodWithImage here?

test/e2e/issue276_test.go
```diff
@@ -0,0 +1,242 @@
...
+	// triggering a Job failure; the back-off time between the runs is
+	// unfortunately exponential and non-configurable, which makes this test
+	// awfully long. Try to at least make it possible to run in parallel with
+	// other tests.
+	err = framework.Global.Client.Create(context.TODO(), gitops, &framework.CleanupOptions{TestContext: ctx, Timeout: timeout * 3, RetryInterval: retryInterval})
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	// Step 2: Wait until Job fails and check if it failed properly
```
akavel:
```suggestion
	// Step 2: Wait until Job fails and check if it printed a clear error message
```

test/e2e/util.go
```diff
@@ -61,6 +61,22 @@ func GetPod(namespace, namePrefix, containsImage string, kubeclient kubernetes.I
...
 
+// GetPodLogs retrieves logs of a given pod
+func GetPodLogs(pod *v1.Pod, kubeclient kubernetes.Interface) (string, error) {
+	req := kubeclient.CoreV1().Pods(pod.Namespace).GetLogs(pod.Name, &v1.PodLogOptions{Timestamps: true})
+	logs, err := req.Stream()
+	if err != nil {
+		return "", xerrors.Errorf("could not get logs for pod %q: %w", pod.Name, err)
+	}
+	buf := &bytes.Buffer{}
+	_, err = io.Copy(buf, logs)
```
akavel: Can we use [io/ioutil.ReadAll](https://golang.org/pkg/io/ioutil/#ReadAll) here?

(I realize I wrote this code originally... I also need my code reviewed :/ )

test/e2e/util.go
```diff
@@ -61,6 +61,22 @@ func GetPod(namespace, namePrefix, containsImage string, kubeclient kubernetes.I
...
+	logs.Close()
```
akavel: The idiomatic way would be to call `defer logs.Close()` just below l.70

test/e2e/util.go
```diff
@@ -61,6 +61,22 @@ func GetPod(namespace, namePrefix, containsImage string, kubeclient kubernetes.I
 	return nil, nil
 }
 
+// GetPodLogs retrieves logs of a given pod
+func GetPodLogs(pod *v1.Pod, kubeclient kubernetes.Interface) (string, error) {
```
akavel: Is it possible to modify DumpJobsLogsOnError to reuse this func?

template-processors/base/bin/processTemplates.sh
```diff
@@ -18,5 +18,11 @@ set -euxo pipefail
 
 echo Processing Templates
 
-# shellcheck disable=SC2086
-cp -R "$CLONED_TEMPLATE_GIT_DIR/"* "$MANIFEST_DIR"/
+# In git, if directory contains no files, it isn't tracked:
+# https://git.wiki.kernel.org/index.php/Git_FAQ#Can_I_add_empty_directories.3F
+if [[ -d "$CLONED_TEMPLATE_GIT_DIR" ]]; then
```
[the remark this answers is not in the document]
sbar95: Great idea, thanks! I was too focused on base.

test/e2e/issue276_test.go
sbar95: The cleanup function deletes the GitOpsConfig after the test. It happens very quickly if there wasn't any fail. But in this case in step 2 we wait for a failed pod to log a proper error message. If the message is logged, the test doesn't fail but the GitOpsConfig still needs to be deleted. So a deletion job is being started - it cannot be done, because there is still a previous creation job pending. And because of our default BackoffLimit value it takes some time. Three minutes is sufficient here. So this test also implicitly tests successul CR deletion after there's a failed creation job - I will make it explicit.

Reference:
https://github.com/operator-framework/operator-sdk/blob/master/doc/test-framework/writing-e2e-tests.md#3-setup-the-test-context-and-resources

> The custom Create function use the controller-runtime's client to create resources and then creates a cleanup function that is called by ctx.Cleanup which deletes the resource and then waits for the resource to be fully deleted before returning.

https://kubernetes.io/docs/concepts/workloads/controllers/jobs-run-to-completion/#pod-backoff-failure-policy

> The back-off limit is set by default to 6. Failed Pods associated with the Job are recreated by the Job controller with an exponential back-off delay (10s, 20s, 40s …) capped at six minutes.

test/e2e/issue276_test.go
sbar95: I'll add a description here.
The test fails with the previous code because the Poll function reaches timeout at some point and an error is set:
https://github.com/kubernetes/apimachinery/blob/731dcecc205498f52a21b12e311af095efb4b188/pkg/util/wait/wait.go#L421-L422
(the WaitFor function is used by pollInternal which is used by Poll)

test/e2e/issue276_test.go
sbar95: WaitForPodWithImage returns true when the pod is Running, I want the pod to be in Succeeded state in this case.

test/e2e/issue276_test.go
```diff
@@ -0,0 +1,276 @@
...
+
+	t.Logf("Deleting CR")
+	err = framework.Global.Client.Delete(context.TODO(), gitops)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	// Three minutes timeout.
+	// FIXME: as of writing this test, "backoffLimit" in job.yaml is set to 4,
+	// which means we need to wait until 5 Pod retries fail, eventually
```
akavel:
```suggestion
	// which means eunomia will wait to launch deletion Job until 5 Pod retries fail, eventually
```

test/e2e/issue276_test.go
```diff
@@ -0,0 +1,276 @@
...
+	// triggering a Job failure; the back-off time between the runs is
```
akavel:
```suggestion
	// triggering the original Job's failure; the back-off time between the runs is
```

template-processors/base/bin/resourceManager.sh
```diff
@@ -80,6 +80,13 @@ function deleteByOldLabels() {
 function createUpdateResources() {
     local owner="$1"
     local timestamp="$(date +%s)"
+    # Check if directory contains only hidden files like .gitkeep, or .gitignore.
+    # This would mean that user purposefully wanted to track an empty directory in git.
+    # https://git.wiki.kernel.org/index.php/Git_FAQ#Can_I_add_empty_directories.3F
+    if [[ -z $(ls "${MANIFEST_DIR}") ]]; then
```
vinny-sabatini: I know the original issue made mention of an empty directory, but I'm wondering if maybe we want to check for a directory without any json/yaml/yml files:
```
$ ls randomfiles/
README.md  sa.yaml
$ kubectl apply -R -f randomfiles/
serviceaccount/my-service-account created
$ rm randomfiles/sa.yaml
$ ls randomfiles/
README.md
$ kubectl apply -R -f randomfiles/
error: error reading [randomfiles/]: recognized file extensions are [.json .yaml .yml]
$ echo $?
1
```

sbar95: While doing this PR I was also thinking about it at some point. In my opinion when something goes wrong with `kubectl apply -R -f`, the error message is descriptive enough for the user to know what might've gone wrong. E.g. in the example you provided the error message says that there were no json, yaml, or yml files and it's rather clear, I think. We could catch an error and wrap it in our own message but in the end the original error message from `kubectl` tells us the most about the problem:
```
$ ls -a broken/
.       ..      sth.yml
$ cat broken/sth.yml 
---
fake fake fake

$ kubectl apply -R -f broken/
error: error validating "broken/sth.yml": error validating data: invalid object to validate; if you choose to ignore these errors, turn validation off with --validate=false
```

vinny-sabatini: I agree, the `kubectl` command does a good job explaining what the error is. So I guess my question would be, rather than just checking if there is nothing in the directory:
`if [[ -z $(ls "${MANIFEST_DIR}") ]]; then`

What if we only applied if yaml/yml/json files exist in the directory:
`if [[ -z $(find ${MANIFEST_DIR} -name "*.yml" -o -name "*.json" -o -name "*.yaml") ]]; then`

If we think that's overkill, I'm ok with how it is. It would just help if a user stubbed in a README or some other file in the directory explaining what it's going to be used for before they added any manifests.

sbar95: I've added a check for yml, yaml, and json files. I created another if block because when directory is empty (has a .gitkeep or other hidden file) we don't want to raise an error (user wanted an empty directory explicitly), and when directory isn't empty but contains no yaml, yml, or json files it's an error.

template-processors/base/bin/resourceManager.sh
```diff
@@ -80,6 +80,16 @@ function deleteByOldLabels() {
...
+        echo "Manifest directory empty, skipping"
+        return
+    elif [[ -z $(find "$MANIFEST_DIR" -regextype posix-extended -iregex '.*\.(ya?ml|json)') ]]; then
```
akavel: if adding this, could we unify the `find` call between here and l.47, so that both look the same?

sbar95: Thanks!

sbar95:
@seanmalloy I think it's now ready to be merged

PR #315, opened 2020-03-10
Delete cronjobs left after Periodic trigger has been removed from GitOpsConfig

### LINKED ISSUE #279: Removing Periodic trigger does not delete the cronjob

eunomia version: v0.1.1

Yes

$ kubectl version
OCP 3.11
</pre></details>

**What did you do?**

I created a GitOpsConfig with only a `Webhook` trigger. I then added a `Periodic` trigger to the GitOpsConfig. The CronJob was created, and then I removed the `Periodic` trigger from the GitOpsConfig.

Example GitOpsConfig:
```
---
apiVersion: eunomia.kohls.io/v1alpha1
kind: GitOpsConfig
metadata:
  name: namespace-poc
spec:
  templateSource:
    uri: https://github.example.com/Operations/openshift-namespaces
    ref: master
    contextDir: namespaces
    secretRef: eunomia-runner-github
  triggers:
  - type: Webhook
  - type: Periodic
    cron: "0 * * * *"
  serviceAccountRef: eunomia-runner
  templateProcessorImage: docker-registry.default.svc:5000/k-images/eunomia-base:latest
  resourceHandlingMode: Apply
  resourceDeletionMode: Delete
```

The Kubernetes CronJob to be deleted for the GitOpsConfig.

**What did you see instead?**

The Kubernetes CronJob was not deleted.
The CronJob was not deleted upon restarting the Eunomia Operator.

sbar95:
**Description**

Delete cronjobs left after GitOpsConfig with Periodic trigger has been
updated to not have such trigger. Now if GitOpsConfig gets updated, and
does not contain Periodic trigger, list cronjobs owned by GitOpsConfig
in reconciliation loop, and delete them.

Add note on how cronjob owned by GitOpsConfig is deleted after
GitOpsConfig has been deleted. It is managed by k8s. Do not handle
cronjob deletion explicitly as it cannot block the creation of a
GitOpsConfig delete job.

ALSO:
Handle situation when there is more than one stuck job. Such situation
can occur when there is a Periodic trigger in a GitOpsConfig which has
wrong image specified.
Previously, if there were more than one stuck job, finalizer wouldn't've
removed them.

Fixes #279 

**Type of change**

**Checklist**

- [ ] Documentation updated (not needed in this PR)

### DIFF
--- a/pkg/controller/gitopsconfig/controller.go
+++ b/pkg/controller/gitopsconfig/controller.go
@@ -491,6 +491,32 @@ func (r *Reconciler) removeFinalizer(ctx context.Context, instance *gitopsv1alph
 	return reconcile.Result{}, nil
 }
 
+// ownedCronJobs retrieves all cronjobs in namespace owner.Namespace whose owner is the passed GitOpsConfig.
+func ownedCronJobs(ctx context.Context, kube client.Client, owner *gitopsv1alpha1.GitOpsConfig) ([]batchv1beta1.CronJob, error) {
+	cronJobs := batchv1beta1.CronJobList{}
+	owned := []batchv1beta1.CronJob{}
+	err := kube.List(ctx, &client.ListOptions{
+		Namespace: owner.Namespace,
+	}, &cronJobs)
+	if err != nil {
+		return nil, xerrors.Errorf("failed to list cronjobs in namespace %s: %w", owner.Namespace, err)
+	}
+
+	for _, cronJob := range cronJobs.Items {
+		ownerRefs := cronJob.GetOwnerReferences()
+		for _, ownerRef := range ownerRefs {
+			if *ownerRef.Controller == true &&
+				ownerRef.APIVersion == owner.APIVersion &&
+				ownerRef.Kind == owner.Kind &&
+				ownerRef.Name == owner.ObjectMeta.Name {
+				owned = append(owned, cronJob)
+				break
+			}
+		}
+	}
+	return owned, nil
+}
+
 // ownedJobs retrieves all jobs in namespace owner.Namespace with value of label tagJobOwner equal to owner.Name.
 func ownedJobs(ctx context.Context, kube client.Client, owner *gitopsv1alpha1.GitOpsConfig) ([]batchv1.Job, error) {
 	jobs := batchv1.JobList{}
--- a/pkg/controller/gitopsconfig/controller.go
+++ b/pkg/controller/gitopsconfig/controller.go
@@ -194,8 +194,22 @@ func (r *Reconciler) Reconcile(request reconcile.Request) (reconcile.Result, err
 		if err != nil {
 			reqLogger.Error(err, "error creating the cronjob, continuing...")
 		}
+	} else {
+		// if there are some leftover cronjobs after removing the Periodic trigger, delete them
+		cronJobs, err := ownedCronJobs(context.TODO(), r.client, instance)
+		if err != nil {
+			reqLogger.Error(err, "unable to list cronjobs", "namespace", instance.Namespace)
+			return reconcile.Result{}, xerrors.Errorf("unable to list cronjobs while checking if there are any left after updating GitOpsConfig: %w", err)
+		}
+		for _, cronJob := range cronJobs {
+			err = r.client.Delete(context.TODO(), &cronJob, client.PropagationPolicy(metav1.DeletePropagationBackground))
+			if err != nil {
+				log.Error(err, "Unable to delete leftover cronjob", "instance", instance.Name, "cronjob", cronJob.Name)
+				return reconcile.Result{}, xerrors.Errorf("Unable to delete leftover cronjob %q for %q: %w", cronJob.Name, instance.Name, err)
+			}
+			log.Info("Deleted leftover cronjob", "instance", instance.Name, "cronjob", cronJob.Name)
+		}
 	}
-	// TODO: if Periodic trigger is removed from CR, remove corresponding CronJob
 
 	if ContainsTrigger(instance, "Change") || ContainsTrigger(instance, "Webhook") {
 		reqLogger.Info("Instance has a change or Webhook trigger, creating job", "instance", instance.GetName())
--- a/pkg/controller/gitopsconfig/controller.go
+++ b/pkg/controller/gitopsconfig/controller.go
@@ -432,28 +432,41 @@ func (r *Reconciler) manageDeletion(instance *gitopsv1alpha1.GitOpsConfig) (reco
 		return reconcile.Result{}, xerrors.Errorf("GitOpsConfig finalizer unable to list owned jobs for %q: %w", instance.Name, err)
 	}
 
-	// If exactly 1 job exists, but it's blocked because of bad image, we
-	// assume the GitOpsConfig never managed to successfully deploy, so we can
-	// just delete the job, remove the finalizer, and be done (#216). It may be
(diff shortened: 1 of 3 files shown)


pkg/controller/gitopsconfig/controller.go
```diff
@@ -432,29 +432,47 @@ func (r *Reconciler) manageDeletion(instance *gitopsv1alpha1.GitOpsConfig) (reco
...
-		if status != nil && status.Waiting != nil && containsString(safeReasons, status.Waiting.Reason) {
-			err = r.client.Delete(context.TODO(), &jobs[0], client.PropagationPolicy(metav1.DeletePropagationBackground))
+	// If a job with owner GitOpsConfig exists, but it's blocked because of bad
+	// image, we assume the GitOpsConfig never managed to successfully deploy,
+	// so we can just delete the job, remove the finalizer, and be done (#216).
+	// It may be either action=create or action=delete job.
+	// There can be more than one stuck job - such situation occurs when there
+	// was a GitOpsConfig with a Periodic trigger which created a few jobs. If
+	// there is a job with GitOpsConfig owner, we remove the finalizer just
+	// after removing all stuck jobs - because such job would block a deletion
```
akavel: What about a situation when there are some stuck jobs and some non-stuck jobs? or do we know somehow that this can't happen?

pkg/controller/gitopsconfig/controller.go
```diff
@@ -432,29 +432,47 @@ func (r *Reconciler) manageDeletion(instance *gitopsv1alpha1.GitOpsConfig) (reco
...
+	// job creation. Stuck jobs created by a cronJob don't block a deletion job
+	// creation but we delete them nontheless to clean up.
```
akavel: Can we avoid deleting the jobs created by a CronJob? (Will deleting the CronJob also result in the jobs being automatically deleted?) Would this make our code bigger or smaller (more complex or simpler)? If we could avoid deleting what we don't have to delete in a way that makes the code simpler, it would be awesome. If it increased complexity, I'm ok keeping it as is, I wouldn't want this code to become even more complex than it already is.

If this is equivalent to just moving the loop from l.463-467 higher, I think it might be worth considering. If the block below  just completely ignored CronJob-borne Jobs, it might possibly make the mental context required when reading/analyzing this code somewhat smaller and thus simpler/easier to understand.

test/e2e/issue279_test.go
```diff
@@ -0,0 +1,301 @@
...
+			return false, nil
+		}
+	})
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	// Step 3: Change Periodic trigger to Webhook one in GitOpsConfig, and apply it
+
+	err = framework.Global.Client.Get(context.TODO(), types.NamespacedName{Name: "gitops-issue279", Namespace: namespace}, gitops)
```
akavel: [optional]

```suggestion
	err = framework.Global.Client.Get(context.TODO(), util.NN{Name: "gitops-issue279", Namespace: namespace}, gitops)
```

test/e2e/issue279_test.go
```diff
@@ -0,0 +1,301 @@
...
+	err = framework.Global.Client.Create(context.TODO(), gitops, &framework.CleanupOptions{TestContext: ctx, Timeout: timeout, RetryInterval: retryInterval})
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	// Step 2: Wait until CronJob creates a Job, whose pod succeeds
+
+	// Two minutes timeout to give cronjob enought time to kick off
+	err = wait.Poll(retryInterval, time.Second*120, func() (done bool, err error) {
+		const name = "gitopsconfig-gitops-issue279"
```
akavel: would it work if we added `-` at the end of the name?

```suggestion
		const name = "gitopsconfig-gitops-issue279-"
```

test/e2e/issue279_test.go
```diff
@@ -0,0 +1,301 @@
...
+	err = framework.Global.Client.Update(context.TODO(), gitops)
+	if err != nil {
+		t.Fatal(err)
+	}
+	t.Logf("GitOpsConfig successfully updated")
+
+	// Step 4: Wait for CronJob to be deleted
+
+	err = wait.Poll(retryInterval, timeout, func() (done bool, err error) {
+		const name = "gitopsconfig-gitops-issue279"
```
akavel: similar as above, can we add `-` suffix?

test/e2e/issue279_test.go
```diff
@@ -0,0 +1,301 @@
...
+
+	err = framework.Global.Client.Create(context.TODO(), gitops, &framework.CleanupOptions{TestContext: ctx, Timeout: timeout, RetryInterval: retryInterval})
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	// Step 2: Make sure that CronJob has been created
+
+	err = wait.Poll(retryInterval, timeout, func() (done bool, err error) {
+		const name = "gitopsconfig-gitops-issue279"
```
akavel: similar as above, `-` suffix?

test/e2e/issue279_test.go
```diff
@@ -0,0 +1,301 @@
...
+	t.Logf("Deleting CR")
+	err = framework.Global.Client.Delete(context.TODO(), gitops)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	// Step 4: Wait for CronJob to be deleted
+
+	err = wait.Poll(retryInterval, timeout, func() (done bool, err error) {
+		const name = "gitopsconfig-gitops-issue279"
```
akavel: suffix `-`?

pkg/controller/gitopsconfig/controller.go
sbar95: There can be completed jobs from before, and stuck jobs from current configuration (because of wrong image). In such situation only stuck jobs are deleted, the rest will be garbage-collected.

test/e2e/issue279_test.go
sbar95: For some reason I don't know the cronJob in the second test has the exact name "gitopsconfig-gitops-issue279". If I remove the suffix, test fails.

test/e2e/issue279_test.go
sbar95: For some reason I don't know the cronJob in the second test has the exact name "gitopsconfig-gitops-issue279". If I remove the suffix, test fails.

pkg/controller/gitopsconfig/controller.go
sbar95: I updated the code, all stuck jobs have to be deleted

seanmalloy:
@akavel FYI this PR now has merge conflicts that need to be fixed.

akavel:
So, after some time thinking, my understanding now is that this PR does currently have basically two parts:

 1. One part that is fixing the actual issue described in #279;
 2. An extra part that is an attempt at fixing another, related TODO/issue in finalizer.

In my opinion, the 1st part seems generally correct and after splitting and a tiny stylistic tweak generally is of great quality. Whereas the 2nd part in my opinion bites a much bigger thing; I'd prefer to pull that part back for now, and try to start working on it separately by writing down an issue describing its complex aspects, to give some better space for thinking and decisions. I plan to go on with this split, I believe this will help us approach this in smaller, easier to digest chunks, and also in fact close the actual issue #279.

Per above explanation, I'm closing this in favor of #322, which contains a slightly modified subset of commits from this PR.

PR #354, opened 2020-07-09
Fix unreliable webhook triggers

### LINKED ISSUE #336: Webhook only kicking off one of multiple gitops configs for same repo and branch
0.1.6

eunomia version:
0.1.6
yes

Using oc command to apply the gitopsconfigs
oc v3.11.0+0cbc58b
kubernetes v1.11.0+d4cacc0
Eunomia is running on an Openshift cluster
openshift v3.11.98
kubernetes v1.11.0+d4cacc0

**What did you do?**

I pushed a helm chart with a simple cron job and 4 different parameter config folders to my github branch.
I created 4 gitOpsConfigs with same helm chart as templateSource and different folders for the parameterSource config directory from my branch
the triggers were set to:
  triggers:                                                                        
  - type: Change                                                                   
  - type: Webhook                                                                  
  - cron: 5 * * * *                                                                
    type: Periodic
I ran oc apply -f each of the gitopsconfig yamls

Once the gitopsconfigs were applied. I made a change to suspend each of the cron jobs in the parameter config folders and pushed that to my branch.

all of the gitOpsConfigs  would kick off a job by the webhook to configure each of the cronjobs.

**What did you see instead?**
only the last gitOpsConfig that was created in the namespace kicked off a job by the webhook.

Smiley73:
**Description**

This pull request fixes the issue with unreliable webhook triggers.

Using GenericEvents and a channel produces inconsistent results. Most OSS operators don't implement this functionality. Switching back to creating the gitopsconfig job directly (Which is the same the reconciler would do, except this works consistently).

Fixes #336 

This PR also adds 2 new parameters to `scripts/e2e-test.sh` in order to speed up local testing cycles:
```
-d|--nodeployment Skip the deployment of the operator
-n|--noimages Skip building the images
-r|--run=<test name> Only run the specified Go e2e test. This will skip all others.
```

**Type of change**

**Checklist**

- [X] Unit tests and e2e tests updated
- [X] Documentation updated

### DIFF
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -221,6 +221,7 @@ make test-unit
 # Optional: Set the environment variable $EUNOMIA_TEST_PAUSE to yes (default is no)
 # Optional: Set the environment variable $EUNOMIA_TEST_SKIP_IMAGES to yes to skip building all images
 # Optional: Set the environment variable $EUNOMIA_TEST_SKIP_DEPLOYMENT to yes to skip the Eunomia deployment
+# Optional: Set the environment variable $EUNOMIA_TEST_GO_RUN to the single test you want to run
 
 make test-e2e
 ```
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -219,6 +219,9 @@ make test-unit
 # Optional: Set the environment variable $EUNOMIA_REF to point to a specific git reference for testing
 # Optional: Set the environment variable $EUNOMIA_TEST_ENV to minishift (default is minikube)
 # Optional: Set the environment variable $EUNOMIA_TEST_PAUSE to yes (default is no)
+# Optional: Set the environment variable $EUNOMIA_TEST_SKIP_IMAGES to yes to skip building all images
+# Optional: Set the environment variable $EUNOMIA_TEST_SKIP_DEPLOYMENT to yes to skip the Eunomia deployment
+
 make test-e2e
 ```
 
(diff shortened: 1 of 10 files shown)


DEVELOPMENT.md
```diff
@@ -219,6 +219,10 @@ make test-unit
 # Optional: Set the environment variable $EUNOMIA_REF to point to a specific git reference for testing
 # Optional: Set the environment variable $EUNOMIA_TEST_ENV to minishift (default is minikube)
 # Optional: Set the environment variable $EUNOMIA_TEST_PAUSE to yes (default is no)
+# Optional: Set the environment variable $EUNOMIA_TEST_SKIP_IMAGES to yes to skip building all images
+# Optional: Set the environment variable $EUNOMIA_TEST_SKIP_DEPLOYMENT to yes to skip the Eunomia deployment
+# Optional: Set the environment variable $EUNOMIA_TEST_GO_RUN to the single test you want to run
+
```
jimmyfigiel, line 225: Just curious, why is this done in environment variables vs parameters?

scripts/e2e-test.sh
```diff
@@ -131,6 +138,23 @@ while (("$#")); do
...
+        ;;
+    -r | --run) # run specific go e2e test
+        if [ -n "$2" ] && [ "${2:0:1}" != "-" ]; then
+            EUNOMIA_TEST_GO_RUN=$2
+            shift 2
+        else
+            echo "Error: Argument for $1 is missing" >&2
+            exit 1
+        fi
+        ;;
```
jimmyfigiel, line 157: Curious as to why this is done via environment variables vs managing the settings in the code?

Smiley73: It's about the scope of the variables. Using `export` allows us to be more flexible within the shell script. The env variables never make it to the parent shell, so there's no concern about it changing anything important.
We also need to use environment variables so that we can control how `make` behaves when called from `.travis.yml`

PR #361, opened 2020-07-22
Increase CPU limit to 1 to support heavier template processing

Smiley73:
**Description**

Complex helm charts take forever to process. This change allows the pod to take up to 1 CPU (limit), but leaves the request at 200m.

**Type of change**

**Checklist**

- [ ] Documentation updated

### DIFF
--- a/build/job-templates/cronjob.yaml
+++ b/build/job-templates/cronjob.yaml
@@ -99,7 +99,7 @@ spec:
                 cpu: '200m'
                 memory: '256Mi'
               limits:
-                cpu: '200m'
+                cpu: '1'
                 memory: '256Mi'
           volumes:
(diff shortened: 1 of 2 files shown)


vinny-sabatini:
I'm ok with this, but something to keep in mind is the job request and limits are currently configurable without making the change to Eunomia directly.
Ref: https://github.com/KohlsTechnology/eunomia#job-templates

I believe there was talk of possibly making the requests/limits available in the CRD whenever we re-visit the definition, but the v1alpha2 discussion has been on hold at this point.

Smiley73:
Yeap. In v1alpha2 we want to actually move the resource setting into the CR, to be more flexible. Changing the template is a cluster-wide thing..

Issue #351, opened 2020-06-26
Resource Deletion Incorrectly Triggered


seanmalloy:
**What version of eunomia are you using?**

<details><summary><code>kubectl exec $EUNOMIA_POD curl localhost:8383/metrics</code> Output</summary><br><pre>
$ kubectl get -n eunomia-operator endpoints/eunomia-operator -o jsonpath='{.subsets[*].addresses[*].targetRef.name}' | xargs -I% kubectl exec -n eunomia-operator % -- curl -sS localhost:8383/metrics | grep eunomia_build_info

</pre></details>

eunomia version: v0.1.6

**Does this issue reproduce with the latest release?**
Yes

**What operating system and processor architecture are you using (`kubectl version`)?**

<details><summary><code>kubectl version</code> Output</summary><br><pre>
$ kubectl version
Client Version: version.Info{Major:"1", Minor:"18", GitVersion:"v1.18.3", GitCommit:"2e7996e3e2712684bc73f0dec0200d64eec7fe40", GitTreeState:"clean", BuildDate:"2020-05-20T12:52:00Z", GoVersion:"go1.13.9", Compiler:"gc", Platform:"darwin/amd64"}
Server Version: version.Info{Major:"1", Minor:"16+", GitVersion:"v1.16.8-gke.15", GitCommit:"9cabee15e0922c3b36724de4866a98f6c2da5e6a", GitTreeState:"clean", BuildDate:"2020-05-01T21:47:04Z", GoVersion:"go1.13.8b4", Compiler:"gc", Platform:"linux/amd64"}
</pre></details>

**What did you do?**
Updated the parameters(yaml files) in my git repo. Specifically a block of yaml was commented out from one of the input parameter files this should have caused a single k8s resource to be removed from the target cluster. Instead a large number of k8s resources were deleted from the cluster that we did not expect to be deleted.

The single resource that should have been removed was a [GCP Config Connector ComputeFirewall](https://cloud.google.com/config-connector/docs/reference/resource-docs/compute/computefirewall). The additional resources that were removed were things like ComputeRouters, ComputeSubnetworks, ComputeSharedVPCHostProjects, and ComputeNetworks.

Note we are using a custom proprietary template processor that uses Ansible. This line of code that is provided from the base template processor image is what triggered the deletion: https://github.com/KohlsTechnology/eunomia/blob/988739ae96d923d1d9913ebeff54dabdae942058/template-processors/base/bin/resourceManager.sh#L96

[...]

PR #362, opened 2020-07-24
build: set CronJob concurrencyPolicy to Forbid

vinny-sabatini:
**Description**

To ensure the `CronJob` controller does not unexpectedly spawn multiple
K8S `Job`s, we are setting the `concurrentyPolicy` to `Forbid`.

To ensure the `CronJob` controller does not create a job too far outside
of the defined schedule, the `startingDeadlineSeconds` is being set to 1
minute.

This does not prevent a `CronJob` from creating a `Job` at the same time
that Eunomia would create a `Job`. It appears that is what happened
in issue #351. A future PR will be submitted to try to make the deletion
logic less disruptive in the event that multiple jobs are running at
the same time.

For reference, the Kubernetes API Reference documentation states the
`concurrentyPolicy` does the following:

> Specifies how to treat concurrent executions of a Job. Valid values are:
> - "Allow" (default): allows CronJobs to run concurrently;
> - "Forbid": forbids concurrent runs, skipping next run if previous run hasn't finished yet;
> - "Replace": cancels currently running job and replaces it with a new one

Relates to #351 

**Type of change**

**Checklist**

- [ ] ~~Documentation updated~~

### DIFF
--- a/build/job-templates/cronjob.yaml
+++ b/build/job-templates/cronjob.yaml
@@ -4,6 +4,8 @@ metadata:
   name: gitopsconfig-{{ .Config.ObjectMeta.Name }}
   namespace: {{ .Config.ObjectMeta.Namespace }}
 spec:
+  concurrencyPolicy: Forbid
+  startingDeadlineSeconds: 60
   schedule: "{{ getCron .Config }}"
   jobTemplate:
     metadata:


vinny-sabatini:
My initial thought for testing this would be configuring a `CronJob` with a schedule to run every minute, and then verify no additional `Job`s are created while another one is in progress.
At this point I don't think I want to add that test into the code base because:
- The change is Kubernetes specific functionality, not Eunomia specific functionality
- Users have the ability to replace the `CronJob` configuration
- In the end, we should not have a hard dependency of only one `GitOpsConfig` `Job`/`CronJob` running at the same time. The Eunomia `Job`s needs to be idempotent, especially the resource deletion process.

I am open to implementing a different test for this change, but at this point I do not know what that test would look like.

seanmalloy:
@vinny-sabatini thanks for the detailed commit message. This is greatly appreciated!

PR #364, opened 2020-07-29
template/base: delete resources with older timestamps

### LINKED ISSUE #351: Resource Deletion Incorrectly Triggered

vinny-sabatini:
**Description**

We are only deleting resources if the `gitopsconfig.eunomia.kohls.io/applied`
label is a value less than the current running jobs timestamp
rather than deleting resources if the timestamps are not equal.
This combined with aa704eb should greatly the reduce likelihood
of multiple jobs running at the same time from deleting resources
that are still in Git.

There is still a scenario where if JobA starts before JobB and then JobB
applies the resources before JobA, then JobA would see the timestamp as
old and would delete resources.

Fixes #351

**Type of change**

**Checklist**

- [ ] ~~Documentation updated~~


vinny-sabatini:
I am working on creating a new e2e test that runs two jobs at the same time to confirm resources are not deleted.
I do have one concern with adding it because there is still a race condition that could cause it to fail, however I think it is still a useful test to include to prevent our codebase from regressing.

sanbornick:
So, since this is going this far, when applying resources with the timestamp you could complicate this further by checking current timestamp and if it is newer than the currently be applied timestamp, then not apply the resource as something newer and possibly different already has so this should skip it. With that and what you are doing on deletion check I think we may actually prevent collisions even if the jobs run at different speeds. 🤞

sanbornick:
Actually there could be a race as each run overlaps the read-compute-update/skip loop which then brings in the need for a semaphore but given no shared state the semaphore would have to be something like an annotation based on a unique value/hash by run... but even so since the read lock then set lock process is itself 2 atomic operations a race still exists

template-processors/base/bin/resourceManager.sh
```diff
@@ -76,8 +76,30 @@ function deleteByOldLabels() {
...
+                    local resource_timestamp="$(yq -r -y '.items['"$i"'].metadata.labels."'$TAG_APPLIED'"' /tmp/check_deletion.yaml | head -qn 1)"
+                    if [[ "$resource_timestamp" -lt "$timestamp" ]]; then
+                        local delete_name=$(yq -y '.items['"$i"'].metadata.name' /tmp/check_deletion.yaml | head -qn 1)
+                        local delete_kind=$(yq -y '.items['"$i"'].kind' /tmp/check_deletion.yaml | head -qn 1)
+                        local delete_namespace_exists=$(yq -y '.items['"$i"'].metadata | has("namespace")' /tmp/check_deletion.yaml | head -qn 1)
+                        if [[ "$delete_namespace_exists" =~ "false" ]]; then
+                            kube delete --wait=false "$delete_kind" "$delete_name"
+                        else
+                            local delete_namespace=$(yq -y '.items['"$i"'].metadata.namespace' /tmp/check_deletion.yaml | head -qn 1)
+                            kube delete --wait=false "$delete_kind" "$delete_name" -n "$delete_namespace"
```
Smiley73, line 95: With the settings we currently have in place, won't the whole script fail with an error if one resource can't be deleted for some reason?

vinny-sabatini: Yes that is the case. If we are concerned about that rather than deleting immediately, we could create a file that includes all of the resources that meet the deletion criteria and then run the kubectl delete on the file.

Smiley73: Or just make sure the return code of kubectl is always 0 so it keeps going through the loop and just logs the error

vinny-sabatini: I think it would be better to have pod fail and retry rather than failing silently with a log if something failed to delete.
1) If the job pod fails, the job would create another pod to retry and would hopefully be able to delete the resource that it failed to delete before.
2) If it is just logged, the user may never know if there was a resource that was stuck unless they are actively monitoring the logs of their jobs. I think I would be more open to this option if we had a metric tracking this sort of event.

Therefore my vote would be to keep the code as is or compile the list of resources to delete and run a single delete API call to remove all of the necessary resources at one time.

vinny-sabatini:
I have created an e2e test for this PR, I believe it is ready for final review.

template-processors/base/bin/resourceManager.sh
Smiley73: FINE 👍 

Yeah this makes sense. The user should be aware of the failure and not assume it successfully deleted. 
One discussion we could have is if we want to rely on the job restarting and retrying, or if that logic should be in the script itself (e.g. retry 3 times with exponential back-off, then loudly fail). This is how we're handling it in most of our other code we're writing internally.

test/e2e/util.go
```diff
@@ -208,6 +209,21 @@ func GetCronJob(namespace, namePrefix string, kubeclient kubernetes.Interface) (
 	return nil, nil
 }
 
+// GetJob retrieves a given cronJob based on namespace, and the cronJob name prefix
+func GetJob(namespace, namePrefix string, kubeclient kubernetes.Interface) (*batchv1.Job, error) {
```
Smiley73, line 213: Wouldn't it be better to call this function `GetCronJob` ?

test/e2e/issue351_test.go
```diff
@@ -0,0 +1,161 @@
...
+		},
+	}
+
+	err = framework.Global.Client.Create(ctx, gitops, &framework.CleanupOptions{TestContext: ctx.TestCtx, Timeout: timeout, RetryInterval: retryInterval})
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	// When the initial job is created, we will use it as a template to create two additional jobs at the same time
+	err = wait.Poll(retryInterval, 60*time.Second, func() (done bool, err error) {
```
Smiley73, line 78: Can you move this into a function in `util` please? This would come in handy for other tests also.

test/e2e/issue351_test.go
```diff
@@ -0,0 +1,161 @@
...
+		default:
+			t.Logf("Both jobs are not done")
+			return false, nil
+		}
+	})
+	if err != nil {
+		t.Error(err)
+	}
+
+	// Ensure the pod is pod is not deleted when the jobs complete
```
Smiley73, line 156: Please fix comment

test/e2e/issue351_test.go
```diff
@@ -0,0 +1,161 @@
...
+	err = WaitForPodWithImage(t, framework.Global, ctx.namespace, "hello-world-modes", "hello-app:1.0", retryInterval, timeout)
```
Smiley73, line 157: `hello-world-modes` ?

test/e2e/util.go
vinny-sabatini: No this should be GetJob, I accidentally forgot to update the comment to properly explain what the function was doing. I believe it should be good now

test/e2e/issue351_test.go
vinny-sabatini: I updated the test to use a more generic set of test data, the `hello-world-modes` was a result of using the `modes_test.go` as a template for generating this test. I also updated the job to look for the deployment rather than the pod because that's more important from this tests perspective.

vinny-sabatini:
> Did you run the test case without your fix in place and does it catch the problem?

Good call out. I have just ran this test against the code in the master branch and it does catch the problem where multiple jobs at the same result in the deployment being wrongfully deleted.

template-processors/base/bin/resourceManager.sh
sanbornick: given that reconciliation is not in parallel backoff but eventual success on many resources could dramatically increase the time to run if that is a concern.

vinny-sabatini: I believe I would rather have Kubernetes manage the retries for this case.
I don't think I would expect the deletion failure to be that common of an issue. If deletion do start to happen frequent enough that it is an issue, we could re-address that in a different PR.

PR #376, opened 2020-12-03
Remove resource deletion race condition between jobs

### LINKED ISSUE #370: Eunomia will randomly delete objects when multiple jobs for the same CR run concurrently
Just stubbing this in now so we have an issue to track all the stuff going on around this. Will update later.

All versions are affected

Yes

All os/arch are affected

**What did you do?**
Create a CR with multiple triggers such as a cron + webhook

My resources to get managed without issue.

**What did you see instead?**
Some resources are randomly deleted. They will get recreated on a subsequent run, but they were removed in the interim. This can result in loss for things that had state, such as PVC's. When considering deletion chaining, if the object affected is a namespace or a gitopsconfig CR with deletion mode set to delete many resources can be deleted and dramatic outages caused.

**FOR THIS REASON WE DO NOT RECOMMEND RUNNING EUNOMIA IN ANYTHING OTHER THAN EXPERIMENTAL ENVIRONMENTS AS YOU WILL HAVE OUTAGES AND DATA LOSS.**

vinny-sabatini:
**Description**

Utilize the `metadata.resourceVersion` field in k8s manifests to prevent multiple concurrent jobs from overwriting resources. This is being done because the resource deletion logic relies on owner and timestamp labels, and if there is a conflict in the labels because of concurrent jobs, resources could be unexpectedly deleted.

The solution was based on the Kubernetes documentation around concurrency: https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#concurrency-control-and-consistency

Fixes #370 

There are also small ordering tweaks to the `base` and `openshift-provision` templates to reduce image build times while making changes to the scripts in the template processors.

**Type of change**

**Checklist**

- [x] Documentation updated
- [x] Update code to handle single files with multiple resources defined

### DIFF
--- a/template-processors/base/bin/appendResourceVersion.py
+++ b/template-processors/base/bin/appendResourceVersion.py
@@ -1,0 +1,106 @@
+#!/usr/bin/env python3
+
+# Copyright 2020 Kohl's Department Stores, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#       http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import logging
+import os
+from pathlib import Path
+import subprocess
+import sys
+import tempfile
+import yaml
+
+# appendResourceVersion - patches the YAML&JSON files in $MANIFEST_DIR,
+# adding the metadata.resourceVersion for each resource being managed.
+# This is intended to serve as a locking mechanism when applying resources
+# in which Kubernetes will fail the apply with a StatusConflict (HTTP status code 409)
+# Ref https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#concurrency-control-and-consistency
+#
+# Inputs:
+#
+# MANIFEST_DIR environment variable
+
+if __name__ == "__main__":
+    logging.basicConfig(stream=sys.stdout, level=logging.INFO)
+    logging.info("starting appendResourceVersion")
+    manifest_dir = os.getenv('MANIFEST_DIR')
+    with open('/var/run/secrets/kubernetes.io/serviceaccount/token') as x: token = x.read()
+    files = list(Path(manifest_dir).rglob("*.yml")) + list(Path(manifest_dir).rglob("*.yaml")) + list(Path(manifest_dir).rglob("*.json"))
+    for filename in files:
+        logging.info("processing file {}".format(filename))
+        try:
+            data = yaml.safe_load(subprocess.run(["kubectl",
+                "-s",
+                "https://kubernetes.default.svc:443",
+                "--token",
+                token,
+                "--certificate-authority",
+                "/var/run/secrets/kubernetes.io/serviceaccount/ca.crt",
+                "get",
+                "--ignore-not-found",
+                "-f",
+                filename,
+                "-o",
+                "yaml"], stdout=subprocess.PIPE).stdout)
+            if data is None:
+                logging.error("no kubectl get output for file {}".format(filename))
+                continue
+            if "kind" in data and data["kind"] == "List":
+                logging.info("file {} contains a list".format(filename))
+                if "items" not in data:
+                    logging.info("file {} has list with zero items".format(filename))
+                    continue
+                resource_version = {}
+                for item in data["items"]:
+                    if "metadata" in item and "resourceVersion" in item["metadata"]:
+                        gvk_name = item["apiVersion"] + item["kind"] + item["metadata"]["name"]
+                        resource_version[gvk_name] = item["metadata"]["resourceVersion"]
+                        logging.info("got resource version {} for {}".format(item["metadata"]["resourceVersion"], gvk_name))
+                    else:
+                        logging.error("failed to get resource version for file {}".format(filename))
+                        continue
+                with open(filename, 'r+') as stream:
+                    try:
+                        new_docs = []
+                        docs = yaml.safe_load_all(stream)
+                        for doc in docs:
+                            gvk_name = doc["apiVersion"] + doc["kind"] + doc["metadata"]["name"]
+                            if gvk_name in resource_version:
+                                doc["metadata"]["resourceVersion"] = resource_version[gvk_name]
+                            else:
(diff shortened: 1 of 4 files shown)


vinny-sabatini:
@sanbornick / @seanmalloy here is the start of my attempt to fix #370 . Please let me know if you think I'm on the right track or if you think I need to change the general direction of this.

As mentioned in the TODO comment, the code still has to be updated to work with files that contain multiple Kubernetes resources as the `get` request will return a K8S list and would need some logic to properly append the `resourceVersion` to each resource returned.

I also have to attempt to write a test to simulate this race condition...I might need assistance in trying to figure out how I can do that as the window for this is quite small.

sanbornick:
Consistently reproducing is going to be a challenge. There's also several races that need to be tested.

An example for one of the races would be to start two jobs, pause the first.... somehow... then resume it after the second finishes.

seanmalloy:
FYI ... @vinny-sabatini I rebased this PR against the latest code from the master branch, so that we are now using Go 1.15.5. I'll start working on figuring out how to handle multiple YAML objects in a single file.

Then after that we need to figure out how to write tests for this bug.

template-processors/base/bin/resourceManager.sh
```diff
@@ -54,6 +54,41 @@ function addLabels() {
...
+    mkdir -p "$tmpdir/$MANIFEST_DIR"
+    mkdir -p "$tmpdir/processed_files/$MANIFEST_DIR"
+    # shellcheck disable=SC2044
+    for file in $(find "$MANIFEST_DIR" -regextype posix-extended -iregex '.*\.(ya?ml|json)'); do
+        mkdir -p "$tmpdir"/"$(dirname "$file")"
+        mkdir -p "$tmpdir"/processed_files/"$(dirname "$file")"
+        kube get --ignore-not-found -f "$file" -o yaml >"$tmpdir/$file"
+        local kind="$(cat "$tmpdir"/"$file" | yq -y .kind)"
+        if [[ "$kind" =~ "List" ]]; then
+            #TODO: figure out how to handle this
```
seanmalloy, line 73: @vinny-sabatini how were you able to trigger a `kind: List` being returned from the k8s API server? Did the list only container a single element or did it contain multiple elements?


Looking at the other code that we have in this script. We did not currently support multiple YAML objects in a single  file. So if the API server returns list with more than one element than I'm not sure how we would handle that.

My naive solution would be to add something this ....
```
if [ number_of_list_elements != 1 ]; then
    # something is wrong
    exit 1
fi
# now process the first list element
```

seanmalloy: @sanbornick and @seegeekrow feel free to add your thoughts.

vinny-sabatini: I achieved this by retrieving the `hello-world-yaml` resources after I deployed them to the cluster.
```
$ kubectl get -f eunomia/examples/hello-world-yaml/template1/hello-world.yaml
```

seanmalloy: This example is causing a `List` to be returned.
https://github.com/KohlsTechnology/eunomia/blob/master/examples/hello-world-yaml/template1/hello-world.yaml

sanbornick: [This](https://github.com/KohlsTechnology/eunomia/blob/754565064f351610598cd485cc089867377d8cd4/template-processors/base/bin/resourceManager.sh#L47-L54) I assume is applying the annotations to just the first document, which also is a gap in testing as we should check for unintentionally unmanaged objects, for  [this example](https://github.com/KohlsTechnology/eunomia/blob/master/examples/hello-world-yaml/template1/hello-world.yaml). I would just stub it like that for now and open a new 🐛. This needs to support multiple docs as that pattern is ubiquitous in the community.

sanbornick: wrote that a day ago, not sure why it just posted now...

sanbornick:
Seems like the current approach would try a resource once, and if it fails will stop the run. This would possibly prevent jobs from ever completing if they are managing a frequently updated resource such as one with an aggressive operator updating the status. Should it retry by getting the resource again? If that was done an additional short circuit could be added to stop a job if it re-gets a resource and encounters an applied timestamp from a more recent job.

Smiley73:
@seanmalloy why python? Wouldn't golang be better? The new go-yq v4 might also have what you need, did you check that out?

seanmalloy:
> @seanmalloy why python? Wouldn't golang be better?

To fix this bug I believe we have out grown shell scripting capabilities. I needed something more than just a shell script to solve this problem. Without re-writing all the shell scripts in a new language I felt like writing just the code to fix this bug was simplest using Python.

A longer term replacement for all of the shell scripts is required at some point in the future.

> The new go-yq v4 might also have what you need, did you check that out?

No I did not try v4 of https://github.com/mikefarah/yq. I believe the fix required for this bug has out grown shell scripts.

seanmalloy:
> Seems like the current approach would try a resource once, and if it fails will stop the run. This would possibly prevent jobs from ever completing if they are managing a frequently updated resource such as one with an aggressive operator updating the status. Should it retry by getting the resource again? If that was done an additional short circuit could be added to stop a job if it re-gets a resource and encounters an applied timestamp from a more recent job.

This change has not been made yet. Also, not tests have been written. Other than these two things I believe the fix is working.

I think this might require re-writing more of `resourceManager.sh` in something other than a shell script. Ideally we would get the HTTP return code 409(StatusConflict) and only then do the retry. Also, label comparisons would probably mean more temporary files if it is done as a shell script.

seanmalloy:
@vinny-sabatini please review my changes and let me know what you think.

template-processors/base/bin/appendResourceVersion.py
```diff
@@ -0,0 +1,106 @@
...
+                "-f",
+                filename,
+                "-o",
+                "yaml"], stdout=subprocess.PIPE).stdout)
+            if data is None:
+                logging.error("no kubectl get output for file {}".format(filename))
+                continue
+            if "kind" in data and data["kind"] == "List":
+                logging.info("file {} contains a list".format(filename))
+                if "items" not in data:
```
vinny-sabatini, line 62: I think `items` would exist in the yaml, but it would be an empty list (example below), so we would want to check that `items` exists and its length is 0.
```
apiVersion: v1
items: []
kind: List
metadata:
  resourceVersion: ""
  selfLink: ""
```

```suggestion
                if "items" in data and len(data["items"]) == 0:
```

vinny-sabatini (commented):
@seanmalloy this looks good to me other than the one comment.

On another note, since I opened the PR, I am not able to approve the PR, so someone else will have to take care of that.

vinny-sabatini (commented):
LGTM

seanmalloy (approved):

PR #380, opened 2021-01-11
Replace shell script with hierarchy binary

### LINKED ISSUE #79: Replace template image shell scripts with golang program
We currently use a couple of shell scripts to initialize the templating engines and do some other logic. We need to check if it makes sense to replace all of that logic with a golang program, so that we don't have such bloated images. We just have to be careful not to add a lot of additional work and not to distract from the main operator work.

## What happens?
- Large images due to shell scripts

- We have tiny and efficient images

Smiley73:
**Description**

Replace the parameter processing shell script with a more solid golang program `hierarchy`, which has been released as a separate repo: https://github.com/KohlsTechnology/hierarchy
This allows also for better debugging of hierarchy issue.

Part of #79 

**Type of change**

**Checklist**

- [X] Documentation updated


25region (commented):
LGTM

MeganMoran (dismissed):

MeganMoran (approved):

